PR #1, opened 2023-04-18
First version of the UPF operator

gruyaume:
First version of the UPF operator

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,97 @@
+name: CI
+
+on:
+  push:
+
+jobs:
+  lint-report:
+    name: Lint report
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install tox
+        run: pip install tox
+      - name: Run tests using tox
+        run: tox -e lint
+
+  static-analysis:
+    name: Static analysis
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install tox
+        run: pip install tox
+      - name: Run tests using tox
+        run: tox -e static
+
+  unit-tests-with-coverage:
+    name: Unit tests
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install tox
+        run: pip install tox
+      - name: Run tests using tox
+        run: tox -e unit
+  
+  integration-test:
+    name: Integration tests
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup LXD
+        uses: canonical/setup-lxd@main
+        with:
+          channel: 5.12/stable
+      - name: Setup operator environment
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+          provider: microk8s
+      - name: Run integration tests
+        run: tox -e integration
+      - name: Archive Tested Charm
+        uses: actions/upload-artifact@v3
+        if: ${{ github.ref_name == 'main' }}
+        with:
+          name: tested-charm
+          path: .tox/**/charmed-5g-upf_ubuntu-22.04-amd64.charm
+          retention-days: 5
+      - name: Archive charmcraft logs
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: charmcraft-logs
+          path: /home/runner/.local/state/charmcraft/log/*.log
+      - name: Archive juju crashdump
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: juju-crashdump
+          path: juju-crashdump-*.tar.xz
+
+  publish-charm:
+    name: Publish Charm
+    needs: integration-test
+    runs-on: ubuntu-22.04
+    if: ${{ github.ref_name == 'main' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install charmcraft
+        run: sudo snap install charmcraft --classic
+      - name: Fetch Tested Charm
+        uses: actions/download-artifact@v3
+        with:
+          name: tested-charm
+      - name: Move charm in current directory
+        run: find ./ -name charmed-5g-upf_ubuntu-22.04-amd64.charm -exec mv -t ./ {} \;
+      - name: Select Charmhub channel
+        uses: canonical/charming-actions/channel@2.2.5
+        id: channel
+      - name: Upload charm to Charmhub
+        uses: canonical/charming-actions/upload-charm@2.2.5
+        with:
+          credentials: "${{ secrets.CHARMCRAFT_AUTH }}"
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+          channel: "${{ steps.channel.outputs.name }}"
--- a/README.md
+++ b/README.md
@@ -1,1 +1,21 @@
 # charmed-5g-upf-operator
+
+Charmed Operator for Charmed 5G's User Plane Function (UPF).
+
+
+## Pre-requisites
+
+Kubernetes cluster with Multus.
+
+## Usage
+
+```bash
+juju deploy upf-operator --trust --channel=edge
+```
+
+## Image
+
+- **bessd**: omecproject/upf-epc-bess:master-d1af749
+- **routectl**: omecproject/upf-epc-bess:master-d1af749
+- **web**: omecproject/upf-epc-bess:master-d1af749
+- **pfcp-agent**: omecproject/upf-epc-pfcpiface:master-d1af749
--- a/src/charm.py
+++ b/src/charm.py
@@ -1,0 +1,337 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Charmed operator for the 5G UPF service."""
+
+import logging
+from typing import Optional
+
+from charms.observability_libs.v1.kubernetes_service_patch import (  # type: ignore[import]  # noqa: E501
+    KubernetesServicePatch,
+)
+from charms.prometheus_k8s.v0.prometheus_scrape import (  # type: ignore[import]  # noqa: E501
+    MetricsEndpointProvider,
+)
+from jinja2 import Environment, FileSystemLoader
+from lightkube.models.core_v1 import ServicePort
+from ops.charm import CharmBase, InstallEvent, PebbleReadyEvent, RemoveEvent
+from ops.main import main
+from ops.model import ActiveStatus, Container, ModelError, WaitingStatus
+from ops.pebble import ExecError, Layer
+
+from kubernetes import Kubernetes
+
+logger = logging.getLogger(__name__)
+
+BESSD_CONTAINER_CONFIG_PATH = "/etc/bess/conf"
+PFCP_AGENT_CONTAINER_CONFIG_PATH = "/tmp/conf"
+CONFIG_FILE_NAME = "upf.json"
+PROMETHEUS_PORT = 8080
+
+
+class UPFOperatorCharm(CharmBase):
+    """Main class to describe juju event handling for the 5G UPF operator."""
+
+    def __init__(self, *args):
+        super().__init__(*args)
+        self._kubernetes = Kubernetes(namespace=self.model.name)
+        self._bessd_container_name = self._bessd_service_name = "bessd"
+        self._routectl_container_name = self._routectl_service_name = "routectl"
+        self._web_container_name = self._web_service_name = "web"
+        self._pfcp_agent_container_name = self._pfcp_agent_service_name = "pfcp-agent"
+        self._bessd_container = self.unit.get_container(self._bessd_container_name)
+        self._routectl_container = self.unit.get_container(self._routectl_container_name)
+        self._web_container = self.unit.get_container(self._web_container_name)
+        self._pfcp_agent_container = self.unit.get_container(self._pfcp_agent_container_name)
+        self.framework.observe(self.on.install, self._on_install)
+        self.framework.observe(self.on.remove, self._on_remove)
+        self.framework.observe(self.on.bessd_pebble_ready, self._on_bessd_pebble_ready)
+        self.framework.observe(self.on.routectl_pebble_ready, self._on_routectl_pebble_ready)
+        self.framework.observe(self.on.web_pebble_ready, self._on_web_pebble_ready)
+        self.framework.observe(self.on.pfcp_agent_pebble_ready, self._on_pfcp_agent_pebble_ready)
+        self._metrics_endpoint = MetricsEndpointProvider(
+            self,
+            jobs=[
+                {
+                    "static_configs": [{"targets": [f"*:{PROMETHEUS_PORT}"]}],
+                }
+            ],
+        )
+        self._service_patcher = KubernetesServicePatch(
+            charm=self,
+            ports=[
+                ServicePort(name="pfcp", port=8805, protocol="UDP"),
+                ServicePort(name="bess-web", port=8000),
+                ServicePort(name="prometheus-exporter", port=PROMETHEUS_PORT),
+            ],
+        )
+
+    def _on_install(self, event: InstallEvent) -> None:
+        self._kubernetes.create_network_attachment_definitions()
+        self._kubernetes.patch_statefulset(statefulset_name=self.app.name)
+        if not self._bessd_container.can_connect():
+            self.unit.status = WaitingStatus("Waiting for bessd container to be ready")
+            event.defer()
+            return
+        if not self._bessd_config_file_is_written:
+            self._write_bessd_config_file()
+
+    def _on_remove(self, event: RemoveEvent) -> None:
+        """Handle remove event."""
+        self._kubernetes.delete_network_attachment_definitions()
+
+    def _write_bessd_config_file(self) -> None:
+        """Write the configuration file for the 5G UPF service."""
+        jinja2_environment = Environment(loader=FileSystemLoader("src/templates/"))
+        template = jinja2_environment.get_template(f"{CONFIG_FILE_NAME}.j2")
+        content = template.render(
+            upf_hostname=self._upf_hostname,
+            mode="af_packet",
+        )
+        self._bessd_container.push(
+            path=f"{BESSD_CONTAINER_CONFIG_PATH}/{CONFIG_FILE_NAME}", source=content
+        )
+        logger.info(f"Pushed {CONFIG_FILE_NAME} config file")
+
+    @property
+    def _upf_hostname(self) -> str:
+        return f"{self.model.app.name}.{self.model.name}.svc.cluster.local"
+
+    @property
+    def _bessd_config_file_is_written(self) -> bool:
+        """Returns whether the bessd config file was written to the workload container."""
+        if not self._bessd_container.exists(f"{BESSD_CONTAINER_CONFIG_PATH}/{CONFIG_FILE_NAME}"):
+            logger.info(f"Config file is not written: {CONFIG_FILE_NAME}")
+            return False
+        logger.info("Config file is written")
+        return True
+
+    @property
+    def _pfcp_agent_config_file_is_written(self) -> bool:
+        """Returns whether the pfcp agent config file was written to the workload container."""
+        if not self._pfcp_agent_container.exists(
+            f"{PFCP_AGENT_CONTAINER_CONFIG_PATH}/{CONFIG_FILE_NAME}"
+        ):
+            logger.info(f"Config file is not written: {CONFIG_FILE_NAME}")
+            return False
+        logger.info("Config file is written")
+        return True
+
+    def _on_bessd_pebble_ready(self, event: PebbleReadyEvent) -> None:
+        """Handle Pebble ready event."""
+        if not self._bessd_config_file_is_written:
+            self.unit.status = WaitingStatus("Waiting for bessd config file to be written")
+            return
+        if not self._kubernetes.statefulset_is_patched(statefulset_name=self.app.name):
+            self.unit.status = WaitingStatus("Waiting for statefulset to be patched")
+            event.defer()
+            return
+        try:
+            self._exec_command_in_bessd_workload(
+                command=["ip", "route", "replace", "192.168.251.0/24", "via", "192.168.252.1"]
+            )
+            self._exec_command_in_bessd_workload(
+                command=[
+                    "ip",
+                    "route",
+                    "replace",
+                    "default",
+                    "via",
+                    "192.168.250.1",
+                    "metric",
+                    "110",
+                ],
+            )
+            self._exec_command_in_bessd_workload(
+                command=[
+                    "iptables",
+                    "-I",
+                    "OUTPUT",
+                    "-p",
+                    "icmp",
+                    "--icmp-type",
+                    "port-unreachable",
+                    "-j",
+                    "DROP",
+                ],
+            )
+        except ExecError:
+            self.unit.status = WaitingStatus("Waiting to be able to prepare bessd container")
+            event.defer()
+            return
+        self._bessd_container.add_layer("upf", self._bessd_pebble_layer, combine=True)
+        self._bessd_container.replan()
+        self._exec_command_in_bessd_workload(
+            command=["bessctl run /opt/bess/bessctl/conf/up4"],
+            environment=self._bessd_environment_variables,
+        )
+        self._set_unit_status()
+
+    def _exec_command_in_bessd_workload(
+        self, command: list, environment: Optional[dict] = None
+    ) -> None:
+        process = self._bessd_container.exec(command=command, timeout=30, environment=environment)
+        try:
+            process.wait_output()
+        except ExecError as e:
+            logger.error("Exited with code %d. Stderr:", e.exit_code)
+            if e.stderr:
+                for line in e.stderr.splitlines():
+                    logger.error("    %s", line)
+            raise e
+
+    def _on_routectl_pebble_ready(self, event: PebbleReadyEvent) -> None:
+        """Handle routectl Pebble ready event."""
+        if not self._kubernetes.statefulset_is_patched(statefulset_name=self.app.name):
+            self.unit.status = WaitingStatus("Waiting for statefulset to be patched")
+            event.defer()
+            return
+        self._routectl_container.add_layer("routectl", self._routectl_pebble_layer, combine=True)
+        self._routectl_container.replan()
+        self._set_unit_status()
+
+    def _on_web_pebble_ready(self, event: PebbleReadyEvent) -> None:
+        """Handle web Pebble ready event."""
+        if not self._kubernetes.statefulset_is_patched(statefulset_name=self.app.name):
+            self.unit.status = WaitingStatus("Waiting for statefulset to be patched")
+            event.defer()
+            return
+        self._web_container.add_layer("web", self._web_pebble_layer, combine=True)
+        self._web_container.replan()
+        self._set_unit_status()
+
+    def _on_pfcp_agent_pebble_ready(self, event: PebbleReadyEvent) -> None:
+        """Handle pfcp agent Pebble ready event."""
+        if not self._pfcp_agent_config_file_is_written:
+            self.unit.status = WaitingStatus("Waiting for pfcp agent config file to be written")
+            event.defer()
+            return
+        if not self._service_is_running(self._bessd_container, self._bessd_service_name):
+            self.unit.status = WaitingStatus("Waiting for bessd service to be running")
+            event.defer()
+            return
+        if not self._kubernetes.statefulset_is_patched(statefulset_name=self.app.name):
+            self.unit.status = WaitingStatus("Waiting for statefulset to be patched")
+            event.defer()
+            return
+        self._pfcp_agent_container.add_layer("pfcp", self._pfcp_agent_pebble_layer, combine=True)
+        self._pfcp_agent_container.replan()
+        self._set_unit_status()
+
+    def _set_unit_status(self) -> None:
+        """Set the application status based on container services being running."""
+        if not self._service_is_running(self._bessd_container, self._bessd_service_name):
+            self.unit.status = WaitingStatus("Waiting for bessd service to run")
+            return
+        if not self._service_is_running(self._routectl_container, self._routectl_service_name):
+            self.unit.status = WaitingStatus("Waiting for routectl service to run")
+            return
+        if not self._service_is_running(self._web_container, self._web_service_name):
+            self.unit.status = WaitingStatus("Waiting for web service to run")
+            return
+        if not self._service_is_running(self._pfcp_agent_container, self._pfcp_agent_service_name):
+            self.unit.status = WaitingStatus("Waiting for pfcp agent service to run")
+            return
+        self.unit.status = ActiveStatus()
+
+    @staticmethod
+    def _service_is_running(container: Container, service_name: str) -> bool:
+        """Returns whether a given service is running."""
+        if not container.can_connect():
+            return False
+        try:
+            service = container.get_service(service_name)
+        except ModelError:
+            return False
+        return service.is_running()
+
+    @property
+    def _bessd_pebble_layer(self) -> Layer:
+        """Returns pebble layer for the bessd container.
+
+        Returns:
+            Layer: Pebble Layer
+        """
+        return Layer(
+            {
+                "summary": "bessd layer",
+                "description": "pebble config layer for bessd",
+                "services": {
+                    self._bessd_service_name: {
+                        "override": "replace",
+                        "startup": "enabled",
+                        "command": "bessd -f -grpc-url=0.0.0.0:10514 -m 0",  # "-m 0" means that we are not using hugepages  # noqa: E501
+                        "environment": self._bessd_environment_variables,
+                    },
+                },
+            }
+        )
+
+    @property
+    def _routectl_pebble_layer(self) -> Layer:
+        """Returns pebble layer for the routectl container."""
+        return Layer(
+            {
+                "summary": "routectl layer",
+                "description": "pebble config layer for routectl",
+                "services": {
+                    self._routectl_service_name: {
+                        "override": "replace",
+                        "startup": "enabled",
+                        "command": "/opt/bess/bessctl/conf/route_control.py -i access core",
+                        "environment": self._routectl_environment_variables,
+                    },
+                },
+            }
+        )
+
+    @property
+    def _web_pebble_layer(self) -> Layer:
+        """Returns pebble layer for the web container."""
+        return Layer(
+            {
+                "summary": "web layer",
+                "description": "pebble config layer for web",
+                "services": {
+                    self._web_service_name: {
+                        "override": "replace",
+                        "startup": "enabled",
+                        "command": "bessctl http 0.0.0.0 8000",
+                    },
+                },
+            }
+        )
+
+    @property
+    def _pfcp_agent_pebble_layer(self) -> Layer:
+        """Returns pebble layer for the pfcp agent container."""
+        return Layer(
+            {
+                "summary": "pfcp agent layer",
+                "description": "pebble config layer for pfcp agent",
+                "services": {
+                    self._pfcp_agent_service_name: {
+                        "override": "replace",
+                        "startup": "enabled",
+                        "command": f"pfcpiface -config {PFCP_AGENT_CONTAINER_CONFIG_PATH}/{CONFIG_FILE_NAME}",  # noqa: E501
+                    },
+                },
+            }
+        )
+
+    @property
+    def _bessd_environment_variables(self) -> dict:
+        return {
+            "CONF_FILE": f"{BESSD_CONTAINER_CONFIG_PATH}/{CONFIG_FILE_NAME}",
+        }
+
+    @property
+    def _routectl_environment_variables(self) -> dict:
+        return {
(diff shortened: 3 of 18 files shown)


tests/unit/test_charm.py
```diff
@@ -0,0 +1,294 @@
...
+    ):
+        patch_exists.return_value = True
+
+        self.harness.container_pebble_ready(container_name="bessd")
+
+        patch_push.assert_not_called()
+
+    @patch("ops.model.Container.exec", new=Mock())
+    @patch("ops.model.Container.exists")
+    def test_given_bessd_config_file_is_written_when_bessd_pebble_ready_then_pebble_plan_is_applied(  # noqa: E501
```
sanchezfdezjavier, line 64: Are we testing the plan is applied, or that is the one we expect? The plan could have been applied and be the same, it clould have been already applied and be correct, and it cloud have been applied, changed, and correct. IDK if I'm making my point, probably is just a matter of naming of the test.

tests/unit/test_charm.py
```diff
@@ -0,0 +1,294 @@
...
+                    timeout=30,
+                    environment=None,
+                ),
+                call().wait_output(),
+            ],
+        )
+
+    @patch("ops.model.Container.exec")
+    @patch("ops.model.Container.exists")
+    def test_given_exec_error_when_bessd_pebble_ready_then_status_is_waiting(
```
sanchezfdezjavier, line 140: Is the config file missing from the test name?
```suggestion
    def test_given_config_file_is_exists_and_exec_error_when_bessd_pebble_ready_then_status_is_waiting(
```

src/charm.py
```diff
@@ -0,0 +1,415 @@
...
+
+    @property
+    def access_network_gateway(self) -> Optional[str]:
+        """Access network gateway IP address."""
+        return self.model.config.get("access-gateway")
+
+    @property
+    def gnb_subnet(self) -> Optional[str]:
+        """Gnodeb subnet."""
+        return self.model.config.get("gnb-subnet")
```
sanchezfdezjavier, line 411: Why do we need these props to be public?

test-requirements.txt
```diff
@@ -0,0 +1,14 @@
+black
+coverage[toml]
+flake8-docstrings
+flake8-builtins
+isort
+juju>=3.1
```
Gmerold, line 6: Not 100% sure, but I think that the action that sets up the env for the itests will use Juju 2.9. If that's the same, integration tests in CI will fail.

src/templates/upf.json.j2
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "core": {
+    "ifname": "core"
+  },
+  "cpiface": {
+    "dnn": "internet",
+    "enable_ue_ip_alloc": false,
+    "hostname": "{{ upf_hostname }}",
+    "http_port": "8080",
+    "ue_ip_pool": "172.250.0.0/16"
```
Gmerold, line 13: Don't we want that to be configurable?

src/charm.py
```diff
@@ -0,0 +1,415 @@
...
+        if not self._pfcp_agent_container.exists(
+            f"{PFCP_AGENT_CONTAINER_CONFIG_PATH}/{CONFIG_FILE_NAME}"
+        ):
+            logger.info(f"Config file is not written: {CONFIG_FILE_NAME}")
+            return False
+        logger.info("Config file is written")
+        return True
+
+    def _on_bessd_pebble_ready(self, event: PebbleReadyEvent) -> None:
+        """Handle Pebble ready event for bessd container.
```
Gmerold, line 134: Or `ConfigChangedEvent`.
And since this function acts on different events, I guess it could use a different name.

src/charm.py
```diff
@@ -0,0 +1,415 @@
...
+
+        Args:
+            event: PebbleReadyEvent
+        """
+        if not self._bessd_config_file_is_written:
+            self._write_bessd_config_file()
```
Gmerold, line 140: Shouldn't we check the container's availability before we try to push a file to it?

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Charm Library used to leverage the Multus Kubernetes CNI in charms.
```
ghislainbourgeois, line 4: Shouldn't this be in lib/?

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+from ops.framework import Object
+
+logger = logging.getLogger(__name__)
+
+NetworkAttachmentDefinitionResource = create_namespaced_resource(
+    group="k8s.cni.cncf.io",
+    version="v1",
+    kind="NetworkAttachmentDefinitionResource",
+    plural="network-attachment-definitions",
+)
```
ghislainbourgeois, line 63: Am I right in assuming this will create K8s resources on import? I would avoid that and probably move it in `__init__`.

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+    def delete_network_attachment_definition(self, name: str) -> None:
+        """Deletes network attachment definition based on name.
+
+        Args:
+            name: NetworkAttachmentDefinition name
+        """
+        self.client.delete(
+            res=NetworkAttachmentDefinitionResource, name=name, namespace=self.namespace
+        )
+        logger.info(f"NetworkAttachmentDefinition {name} deleted")
```
ghislainbourgeois, line 173: We should probably handle errors here

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+        """Creates a NetworkAttachmentDefinitionResource.
+
+        Args:
+            network_attachment_definition: NetworkAttachmentDefinition object
+        """
+        network_attachment_definition_obj = NetworkAttachmentDefinitionResource(
+            metadata=ObjectMeta(name=network_attachment_definition.name),
+            spec=network_attachment_definition.spec,
+        )
+        self.client.create(obj=network_attachment_definition_obj, namespace=self.namespace)  # type: ignore[call-overload]  # noqa: E501
```
ghislainbourgeois, line 159: We should probably handle errors here

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+        #     ]
+        # )
+
+        self.client.patch(
+            res=StatefulSet,
+            name=name,
+            obj=statefulset,
+            patch_type=PatchType.MERGE,
+            namespace=self.namespace,
+        )
```
ghislainbourgeois, line 203: We should probably handle errors here

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+        logger.info(f"Multus annotation added to {name} Statefulset")
+
+    def statefulset_is_patched(
```
ghislainbourgeois, line 206: This would be clearer with the parameters.
```suggestion
    def statefulset_has_annotations(
```

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+        """Returns whether the statefulset has the expected multus annotation.
+
+        Args:
+            name: Statefulset name.
+            network_annotations: List of network annotations
+
+        Returns:
+            bool: Whether the statefulset has the expected multus annotation.
+        """
+        statefulset = self.client.get(res=StatefulSet, name=name, namespace=self.namespace)
```
ghislainbourgeois, line 218: Handle errors here

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+        network_annotations: List[NetworkAnnotation],
+    ):
+        super().__init__(charm, "kubernetes-multus")
+        self.network_attachment_definitions = network_attachment_definitions
+        self.network_annotations = network_annotations
+        self.framework.observe(charm.on.install, self._patch)
+        self.framework.observe(charm.on.upgrade_charm, self._patch)
+        self.framework.observe(charm.on.remove, self._on_remove)
+
+    def _patch(self, event: Union[InstallEvent, UpgradeCharmEvent]) -> None:
```
ghislainbourgeois, line 247:
```suggestion
    def _patch(self, event: ops.framework.EventBase) -> None:
```
We are not using anything on the event, so using type `EventBase` would be better here.

src/charm.py
```diff
@@ -0,0 +1,415 @@
...
+        """
+        if not self._pfcp_agent_container.exists(
+            f"{PFCP_AGENT_CONTAINER_CONFIG_PATH}/{CONFIG_FILE_NAME}"
+        ):
+            logger.info(f"Config file is not written: {CONFIG_FILE_NAME}")
+            return False
+        logger.info("Config file is written")
+        return True
+
+    def _on_bessd_pebble_ready(self, event: PebbleReadyEvent) -> None:
```
ghislainbourgeois, line 133: Unless we require a specific type of event in those handlers, I think we should be more permissive in the type hints and just accept an `EventBase`.

src/charm.py
```diff
@@ -0,0 +1,415 @@
...
+                ],
+            )
+        except ExecError:
+            self.unit.status = WaitingStatus("Cannot execute command in workload, waiting.")
+            event.defer()
+            return
+        self._bessd_container.add_layer("upf", self._bessd_pebble_layer, combine=True)
+        self._bessd_container.replan()
+        self._exec_command_in_bessd_workload(
+            command=["bessctl run /opt/bess/bessctl/conf/up4"],
```
ghislainbourgeois, line 193:
```suggestion
            command=["bessctl run /opt/bess/bessctl/conf/up4".split()],
```
This will separate the tokens and prevent weird shenanigans with quoting

src/charm.py
```diff
@@ -0,0 +1,415 @@
...
+        if not self._service_is_running(self._web_container, self._web_service_name):
+            self.unit.status = WaitingStatus("Waiting for web service to run")
+            return
+        if not self._service_is_running(self._pfcp_agent_container, self._pfcp_agent_service_name):
+            self.unit.status = WaitingStatus("Waiting for pfcp agent service to run")
+            return
+        self.unit.status = ActiveStatus()
+
+    @staticmethod
+    def _service_is_running(container: Container, service_name: str) -> bool:
```
ghislainbourgeois, line 272:
```suggestion
    def _service_is_running_in_container(service_name: str, container: Container) -> bool:
```

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+    dict = asdict
+
+
+@dataclass
+class NetworkAnnotation:
+    """NetworkAnnotation."""
+
+    name: str
+    interface: str
+    ips: Optional[List] = None
```
ghislainbourgeois, line 91:
```suggestion
    ips: Optional[list] = None
```

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+
+        Args:
+            name: NetworkAttachmentDefinition name
+        """
+        self.client.delete(
+            res=NetworkAttachmentDefinitionResource, name=name, namespace=self.namespace
+        )
+        logger.info(f"NetworkAttachmentDefinition {name} deleted")
+
+    def patch_statefulset(self, name: str, network_annotations: List[NetworkAnnotation]) -> None:
```
ghislainbourgeois, line 175:
```suggestion
    def patch_statefulset(self, name: str, network_annotations: list[NetworkAnnotation]) -> None:
```

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+            res=StatefulSet,
+            name=name,
+            obj=statefulset,
+            patch_type=PatchType.MERGE,
+            namespace=self.namespace,
+        )
+        logger.info(f"Multus annotation added to {name} Statefulset")
+
+    def statefulset_is_patched(
+        self, name: str, network_annotations: List[NetworkAnnotation]
```
ghislainbourgeois, line 207:
```suggestion
        self, name: str, network_annotations: list[NetworkAnnotation]
```

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+        return True
+
+
+class KubernetesMultusCharmLib(Object):
+    """Class to be instantiated by charms requiring Multus networking."""
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        network_attachment_definitions: List[NetworkAttachmentDefinition],
```
ghislainbourgeois, line 237:
```suggestion
        network_attachment_definitions: list[NetworkAttachmentDefinition],
```

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+        network_annotations: List[NetworkAnnotation],
```
ghislainbourgeois, line 238:
```suggestion
        network_annotations: list[NetworkAnnotation],
```

src/kubernetes_multus.py
```diff
@@ -0,0 +1,276 @@
...
+                NetworkAnnotation(name="core-net", interface="eth1"),
+            ]
+        )
+```
+"""
+
+import json
+import logging
+from dataclasses import asdict, dataclass, field
+from typing import List, Optional, Union
```
ghislainbourgeois, line 44:
```suggestion
from typing import Optional, Union
```
This is not required with Python 3.10, the default version on 22.04

test-requirements.txt
saltiyazan: I think this is a `pythonlib-juju` version, therefore, the one used here is not a valid version.
If you mean to use `juju 3.1` then the latest `pythonlib-juju` tested with `3.1` would be `3.1.0.1` (you'd need a different version if you use `juju 2.9`)
see more [here](https://github.com/juju/python-libjuju/releases).

.github/workflows/main.yml
```diff
@@ -0,0 +1,97 @@
...
+        with:
+          provider: microk8s
+      - name: Run integration tests
+        run: tox -e integration
+      - name: Archive Tested Charm
+        uses: actions/upload-artifact@v3
+        if: ${{ github.ref_name == 'main' }}
+        with:
+          name: tested-charm
+          path: .tox/**/charmed-5g-upf_ubuntu-22.04-amd64.charm
```
saltiyazan, line 58:
```suggestion
          path: .tox/**/sdcore-upf_ubuntu-22.04-amd64.charm
```

src/charm.py
gruyaume: Fixed

.github/workflows/main.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install charmcraft
+        run: sudo snap install charmcraft --classic
+      - name: Fetch Tested Charm
+        uses: actions/download-artifact@v3
+        with:
+          name: tested-charm
+      - name: Move charm in current directory
+        run: find ./ -name charmed-5g-upf_ubuntu-22.04-amd64.charm -exec mv -t ./ {} \;
```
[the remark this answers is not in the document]
gruyaume: Fixed

src/templates/upf.json.j2
gruyaume: This won't end up being used since we set enable_ue_ip_alloc to False. I simply removed the key/value from the config file now

src/charm.py
```diff
@@ -0,0 +1,415 @@
...
+                    "-p",
+                    "icmp",
+                    "--icmp-type",
+                    "port-unreachable",
+                    "-j",
+                    "DROP",
+                ],
+            )
+        except ExecError:
+            self.unit.status = WaitingStatus("Cannot execute command in workload, waiting.")
```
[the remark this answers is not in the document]
gruyaume: Yeah I changed how that's done now

src/kubernetes_multus.py
gruyaume: Well it should probably be in a lib for a dummy charm for management of k8s resources.

src/kubernetes_multus.py
gruyaume: No this method only creates a class, not an actual kubernetes resource

src/kubernetes_multus.py
gruyaume: Fixed

src/kubernetes_multus.py
gruyaume: Fixed

src/kubernetes_multus.py
gruyaume: Fixed

src/kubernetes_multus.py
gruyaume: Fixed

.github/workflows/main.yml
gruyaume: Fixed

README.md
```diff
@@ -1 +1,21 @@
...
+# sdcore-upf-operator
+
+Charmed Operator for SDCORE's User Plane Function (UPF).
+
+
+## Pre-requisites
+
+Kubernetes cluster with Multus.
+
+## Usage
```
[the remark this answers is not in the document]
gruyaume: I don't think so. On charmhub, there's a tab for config that is generated based on config.yaml

src/kubernetes_multus.py
gruyaume: Fixed

src/kubernetes_multus.py
gruyaume: Fixed

src/kubernetes_multus.py
gruyaume: If the charm fails at creating a network attachment definition it's not gonna work. so right now it would go to error state. What would be a better approach?

src/kubernetes_multus.py
gruyaume: Done

config.yaml
```diff
@@ -0,0 +1,13 @@
...
+    default: 192.168.251.0/24
+    description: gNodeB subnet
+  core-gateway:
+    type: string
+    default: 192.168.250.1
+    description: Core Network Gateway IP address
+  access-gateway:
+    type: string
+    default: 192.168.252.1
+    description: Access Network Gateway IP address
```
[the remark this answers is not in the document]
gruyaume: Done

config.yaml
```diff
@@ -0,0 +1,13 @@
+options:
+  gnb-subnet:
+    type: string
+    default: 192.168.251.0/24
+    description: gNodeB subnet
+  core-gateway:
+    type: string
+    default: 192.168.250.1
+    description: Core Network Gateway IP address
```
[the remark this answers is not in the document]
gruyaume: Done

src/charm.py
gruyaume: Done

README.md
```diff
@@ -1 +1,21 @@
-# charmed-5g-upf-operator
\ No newline at end of file
+# sdcore-upf-operator
+
+Charmed Operator for SDCORE's User Plane Function (UPF).
+
+
```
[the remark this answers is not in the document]
gruyaume: Fixed

src/charm.py
gruyaume: Fixed

src/charm.py
gruyaume: Fixed

src/kubernetes_multus.py
gruyaume: Done

src/kubernetes_multus.py
gruyaume: It's more than annotation, it's also that the container security context is modified

src/charm.py
gruyaume: Fixed by using commas everywhere

README.md
```diff
@@ -1 +1,21 @@
-# charmed-5g-upf-operator
\ No newline at end of file
+# sdcore-upf-operator
+
+Charmed Operator for SDCORE's User Plane Function (UPF).
```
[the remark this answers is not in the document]
gruyaume: I added a bit more, let me know what you think

.github/workflows/main.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Archive juju crashdump
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: juju-crashdump
+          path: juju-crashdump-*.tar.xz
+
+  publish-charm:
+    name: Publish Charm
+    needs: integration-test
```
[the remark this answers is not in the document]
sanchezfdezjavier: I'd split the workflow into separate files as @saltiyazan and I are doing in our PRs [`sdcore-smf`](https://github.com/canonical/sdcore-smf-operator/pull/1/files) [`sdcore-amf`](https://github.com/canonical/sdcore-amf-operator/pull/1/files). It solves the issue @Gmerold is talking about, besides I believe it's cleaner and easier to read.

src/charm.py
```diff
@@ -0,0 +1,553 @@
...
+        return True
+
+    def _configure_bessd_workload(self, event: EventBase) -> None:
+        """Handles events for configuring bessd container.
+
+        Args:
+            event: EventBase
+        """
+        invalid_configs = self._get_invalid_configs()
+        if invalid_configs:
```
sanchezfdezjavier, line 185:
```suggestion
       if invalid_configs := self._get_invalid_configs():
```

src/charm.py
```diff
@@ -0,0 +1,553 @@
...
+    def _write_bessd_config_file(
+        self,
+        upf_hostname: str,
+        upf_mode: str,
+        access_interface_name: str,
+        core_interface_name: str,
+        dnn: str,
+        pod_share_path: str,
+    ) -> None:
+        """Write the configuration file for the 5G UPF service."""
```
sanchezfdezjavier, line 125: Missisng docstring?

src/charm.py
```diff
@@ -0,0 +1,553 @@
...
+
+    @property
+    def _bessd_config_file_is_written(self) -> bool:
+        """Returns whether the bessd config file was written to the workload container.
+
+        Returns:
+            bool: Whether the bessd config file was written
+        """
+        if not self._bessd_container.exists(f"{BESSD_CONTAINER_CONFIG_PATH}/{CONFIG_FILE_NAME}"):
+            logger.info(f"Config file is not written: {CONFIG_FILE_NAME}")
```
sanchezfdezjavier, line 158:
```suggestion
            logger.info("Config file is not written: %s", CONFIG_FILE_NAME)
```

src/charm.py
```diff
@@ -0,0 +1,553 @@
...
+    def _pfcp_agent_config_file_is_written(self) -> bool:
+        """Returns whether the pfcp agent config file was written to the workload container.
+
+        Returns:
+            bool: Whether the pfcp agent config file was written
+        """
+        if not self._pfcp_agent_container.exists(
+            f"{PFCP_AGENT_CONTAINER_CONFIG_PATH}/{CONFIG_FILE_NAME}"
+        ):
+            logger.info(f"Config file is not written: {CONFIG_FILE_NAME}")
```
sanchezfdezjavier, line 173:
```suggestion
            logger.info("Config file is not written: %s", CONFIG_FILE_NAME)
```

src/charm.py
```diff
@@ -0,0 +1,553 @@
...
+        invalid_configs = self._get_invalid_configs()
+        if invalid_configs:
+            self.unit.status = BlockedStatus(
+                f"The following configurations are not valid: {invalid_configs}"
+            )
+            event.defer()
+            return
+        if not self._bessd_container.can_connect():
+            self.unit.status = WaitingStatus("Waiting to be able to connect to the container")
+            event.defer()
```
sanchezfdezjavier, line 193: I think we can remove this `event.defer()` as pebble will trigger. Happens the same in the `smf` as noted by @ghislainbourgeois.

src/charm.py
```diff
@@ -0,0 +1,553 @@
...
+            return
+        self._pfcp_agent_container.add_layer("pfcp", self._pfcp_agent_pebble_layer, combine=True)
+        self._pfcp_agent_container.replan()
+        self._set_unit_status()
+
+    def _set_unit_status(self) -> None:
+        """Set the unit status based on config and container services running."""
+        invalid_configs = self._get_invalid_configs()
+        if invalid_configs:
```
sanchezfdezjavier, line 369:
```suggestion
        if invalid_configs := self._get_invalid_configs():
```

tests/unit/test_charm.py
```diff
@@ -0,0 +1,283 @@
...
+
+        patch_push.assert_has_calls(
+            [
+                call(
+                    path="/etc/bess/conf/upf.json",
+                    source='{\n  "access": {\n    "ifname": "access"\n  },\n  "core": {\n    "ifname": "core"\n  },\n  "cpiface": {\n    "dnn": "internet",\n    "enable_ue_ip_alloc": false,\n    "hostname": "sdcore-upf.whatever.svc.cluster.local",\n    "http_port": "8080"\n  },\n  "enable_notify_bess": true,\n  "gtppsc": true,\n  "hwcksum": true,\n  "log_level": "trace",\n  "max_sessions": 50000,\n  "measure_flow": false,\n  "measure_upf": true,\n  "mode": "af_packet",\n  "notify_sockaddr": "/pod-share/notifycp",\n  "qci_qos_config": [\n    {\n      "burst_duration_ms": 10,\n      "cbs": 50000,\n      "ebs": 50000,\n      "pbs": 50000,\n      "priority": 7,\n      "qci": 0\n    }\n  ],\n  "slice_rate_limit_config": {\n    "n3_bps": 1000000000,\n    "n3_burst_bytes": 12500000,\n    "n6_bps": 1000000000,\n    "n6_burst_bytes": 12500000\n  },\n  "table_sizes": {\n    "appQERLookup": 200000,\n    "farLookup": 150000,\n    "pdrLookup": 50000,\n    "sessionQERLookup": 100000\n  },\n  "workers": 1\n}',  # noqa: E501
```
sanchezfdezjavier, line 42: I'd create an `expected_upf.json` file and load it. It'd be easier to maintain and easier to read.

.github/workflows/main.yml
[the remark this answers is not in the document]
gruyaume: Fixed

src/charm.py
```diff
@@ -0,0 +1,415 @@
...
+
+    @property
+    def _bessd_config_file_is_written(self) -> bool:
+        """Returns whether the bessd config file was written to the workload container.
+
+        Returns:
+            bool: Whether the bessd config file was written
+        """
+        if not self._bessd_container.exists(f"{BESSD_CONTAINER_CONFIG_PATH}/{CONFIG_FILE_NAME}"):
+            logger.info(f"Config file is not written: {CONFIG_FILE_NAME}")
```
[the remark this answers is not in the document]
gruyaume: Agreed, removed

src/charm.py
```diff
@@ -0,0 +1,553 @@
...
+            mode=upf_mode,
+            access_interface_name=access_interface_name,
+            core_interface_name=core_interface_name,
+            dnn=dnn,
+            pod_share_path=pod_share_path,
+        )
+        self._bessd_container.push(
+            path=f"{BESSD_CONTAINER_CONFIG_PATH}/{CONFIG_FILE_NAME}", source=content
+        )
+        logger.info(f"Pushed {CONFIG_FILE_NAME} config file")
```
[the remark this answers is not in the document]
gruyaume: How so?

tests/unit/test_charm.py
gruyaume: Added "expected" in the test name, hopefully it's clearer.

src/charm.py
gruyaume: Done

src/charm.py
gruyaume: Fixed

src/charm.py
gruyaume: Good catch!

src/charm.py
gruyaume: Done

src/kubernetes_multus.py
gruyaume: Fixed

src/kubernetes_multus.py
gruyaume: Ok done, if there's other issues with the multus lib, let's discuss them in the PR for it [here](https://github.com/canonical/kubernetes-charm-libraries/pull/1)

src/kubernetes_multus.py
gruyaume: Done

src/kubernetes_multus.py
gruyaume: Done

src/charm.py
```diff
@@ -0,0 +1,415 @@
...
+                    "-j",
+                    "DROP",
+                ],
+            )
+        except ExecError:
+            self.unit.status = WaitingStatus("Cannot execute command in workload, waiting.")
+            event.defer()
+            return
+        self._bessd_container.add_layer("upf", self._bessd_pebble_layer, combine=True)
+        self._bessd_container.replan()
```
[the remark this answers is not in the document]
gruyaume: Added a restart after the replan

src/charm.py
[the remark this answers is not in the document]
ghislainbourgeois: f-strings are evaluated immediately by the interpreter, so if the log is disabled, it will still do potentially expensive computation. In this case the cost is negligible, but standardizing on one approach for logging prevent some errors like:

```
n = 99999999999
logger.debug(f"Factorial of {n} is {factorial(n)}")
```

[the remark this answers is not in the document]
gruyaume: Hummmm we don't do any computation in logging here, it's always string manipulation. I'd optimise for readability instead of performance, especially in a charm.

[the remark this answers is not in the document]
sanchezfdezjavier: If that's the case should I revert the changes in my PR? I'd strive for having a consistent approach.

src/charm.py
```diff
@@ -290,45 +310,38 @@ def _ip_tables_rule_exists(self) -> bool:
...
         """
-        process = self._bessd_container.exec(command=command, timeout=30, environment=environment)
-        try:
-            return process.wait_output()
-        except ExecError as e:
-            logger.error("Exited with code %d. Stderr:", e.exit_code)
-            if e.stderr:
-                for line in e.stderr.splitlines():
-                    logger.error("    %s", line)
-            raise e
```
sanchezfdezjavier: Why don't we try catch?

gruyaume: This code block doesn't exist anymore

.github/workflows/publish-charm.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Chmod charmcraft logs
+        if: steps.upload.conclusion == 'failure'
+        run: |
+          sudo chmod -R 777 /root/snap/charmcraft/common/cache/charmcraft/log/
+      - name: Archive charmcraft logs
+        if: steps.upload.conclusion == 'failure'
+        uses: actions/upload-artifact@v3
+        with:
+          name: charmcraft-upload-logs
+          path: /root/snap/charmcraft/common/cache/charmcraft/log/*.log
```
ghislainbourgeois, line 36: I don't think this is required when we use `charmcraft upload`, as the output is directly to the console.

src/charm.py
```diff
@@ -0,0 +1,581 @@
...
+        )
+        logger.info(f"Pushed {CONFIG_FILE_NAME} config file")
+
+    @property
+    def _upf_hostname(self) -> str:
+        """Returns the UPF hostname.
+
+        Returns:
+            str: UPF Hostname
+        """
```
ghislainbourgeois, line 157: I have seen a lot of docstrings like this in recent reviews, and I would like to propose that we get rid of them for private properties that are that simple. The name of the property already says everything we need to know.

I would keep some docstrings for anything public, or when the docstring provides more details than the method signature. This is not the case here.

@Gmerold @saltiyazan @sanchezfdezjavier I think this is more of a team standard than anything, and therefore would like your input also.

src/charm.py
```diff
@@ -0,0 +1,581 @@
...
+        while time.time() - initial_time <= timeout:
+            try:
+                self._exec_command_in_bessd_workload(
+                    command="bessctl run /opt/bess/bessctl/conf/up4",
+                    environment=self._bessd_environment_variables,
+                )
+                return
+            except ExecError:
+                logger.info("Failed running configuration for bess")
+                time.sleep(2)
```
ghislainbourgeois, line 244:
```suggestion
        for _ in range(15):
            try:
                self._exec_command_in_bessd_workload(
                    command="bessctl run /opt/bess/bessctl/conf/up4",
                    environment=self._bessd_environment_variables,
                )
                return
            except ExecError:
                logger.info("Failed running configuration for bess")
                time.sleep(2)
```

This is a lot simpler. If you want something more explicit, I think it would be worth it to consider this library: https://github.com/jd/tenacity.

gruyaume: I don't know about the "a lot simpler" and the behavior is different. We used to have an explicit 30 seconds timeout, now we have a 15 tries and no "time" constraint. Ex. if `_exec_command_in_bessd_workload` takes 10 seconds each time, it makes for a 180 second loop (15*(10+2)).

src/charm.py
gruyaume: I agree

.github/workflows/publish-charm.yaml
gruyaume: Removed

src/charm.py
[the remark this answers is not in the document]
gruyaume: I'm not sure I understand the consistency issue, if we use f-string everywhere we keep things consistent. 

Optimising for performance should be done when we have performance issues. Else, we should optimise for code simplicity and readability. Here we are taking a readability hit and optimising for a problem we don't have. If we truly want to improve performance in this charm code, we should start by checking where most of our computing/memory costs are. I expect those not to be in logging right now.

ghislainbourgeois (approved):

src/charm.py
```diff
@@ -0,0 +1,576 @@
...
+        self.framework.observe(self.on.routectl_pebble_ready, self._configure)
+        self.framework.observe(self.on.web_pebble_ready, self._configure)
+        self.framework.observe(self.on.pfcp_agent_pebble_ready, self._configure)
+        self.framework.observe(self.on.config_changed, self._configure)
+        self.framework.observe(self.on.update_status, self._update_status)
+
+    def _update_status(self, event: UpdateStatusEvent):
+        """Sets unit status based on current state."""
+        if self.unit.status != ActiveStatus():
+            self._configure(event)
```
ghislainbourgeois, line 111: This is interesting, was this required to fix the issue you had before?

It seems potentially wasteful, and with all the potential `defer`, couldn't this pile up a bunch of deferred events?

src/charm.py
```diff
@@ -0,0 +1,576 @@
...
+            return
+        if not self._kubernetes_multus.multus_is_configured():
+            self.unit.status = WaitingStatus("Waiting for statefulset to be patched")
+            event.defer()
+            return
+        if not self._bessd_container.can_connect():
+            self.unit.status = WaitingStatus(
+                "Waiting to be able to connect to the `bessd` container"
+            )
+            event.defer()
```
ghislainbourgeois, line 227: I see a lot of `defer` are back, wouldn't this be called again when we receive `PebbleReady`?

I assume this is related to the issue you had with crashes and I am curious to understand a bit more.

ghislainbourgeois (commented):
Looks good overall, but I have some questions.

src/charm.py
gruyaume: No, this wasn't needed and you're right, we may want to remove it. It was more of a "protection". I'll remove it and we can add something along those lines if we really need it in the future.

CONTRIBUTING.md
```diff
@@ -0,0 +1,21 @@
+# Contributing
```
sanchezfdezjavier, line 1: I think we should have the same contributing file across charms. [This is the one for the `smf`](https://github.com/canonical/sdcore-smf-operator/blob/08f720ed7c26042bedcf1c5a72a98836436054aa/CONTRIBUTING.md) based on the `nrf` one but adding the `tox -e fmt`.

test-requirements.txt
```diff
@@ -3,7 +3,7 @@ coverage[toml]
 flake8-docstrings
 flake8-builtins
 isort
-juju==2.9.42.2
+juju==3.1.2.0
```
sanchezfdezjavier, line 6: Why are we pinning the version?

sanchezfdezjavier (approved):
Aside from the nitpicking LGTM

test-requirements.txt
gruyaume: Becaues pythonlibjuju doesn't guarantee compatibility. We pin the version of Juju itself in the integration tests. If we don't pin the `pyhtonlib-juju` version, it's very possible that the 2 won't be compatible very soon.

CONTRIBUTING.md
gruyaume: Done

.github/workflows/push.yaml
```diff
@@ -22,3 +22,5 @@ jobs:
       [lint-report, static-analysis, unit-tests-with-coverage, integration-test]
     if: ${{ github.ref_name == 'main' }}
     uses: ./.github/workflows/publish-charm.yaml
+    secrets:
```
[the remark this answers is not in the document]
gruyaume: Done

README.md
```diff
@@ -1 +1,35 @@
-# charmed-5g-upf-operator
\ No newline at end of file
+# sdcore-upf-operator
```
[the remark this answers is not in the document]
gruyaume: No it's for charmhub

src/charm.py
ghislainbourgeois: Ok, thank you!

ghislainbourgeois (approved):

saltiyazan (approved):

PR #20, opened 2023-07-18
Removes defer calls in config changed handlers

saltiyazan:
# Description

If the container is down there is no need to `defer` config changed event and rerun the handler, pebble_ready event will run the handler when the container is up again.
The main reason for this pr is to avoid running `up4` twice in a row quickly. 

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -217,11 +217,9 @@ class UPFOperatorCharm(CharmBase):
             return
         if not self._bessd_container.can_connect():
             self.unit.status = WaitingStatus("Waiting for bessd container to be ready")
-            event.defer()
             return
         if not self._pfcp_agent_container.can_connect():
             self.unit.status = WaitingStatus("Waiting for pfcp agent container to be ready")
-            event.defer()
             return
         self._on_bessd_pebble_ready(event)
         self._on_pfcp_agent_pebble_ready(event)
--- a/src/charm.py
+++ b/src/charm.py
@@ -228,6 +228,10 @@ class UPFOperatorCharm(CharmBase):
         """Handle Pebble ready event."""
         if not self.unit.is_leader():
             return
+        if not self._bessd_container.can_connect():
+            self.unit.status = WaitingStatus("Waiting for bessd container to be ready")
+            event.defer()
+            return
         if not self._kubernetes_multus.is_ready():
             self.unit.status = WaitingStatus("Waiting for Multus to be ready")
             event.defer()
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -217,11 +217,9 @@ def _on_config_changed(self, event: EventBase):
             return
         if not self._bessd_container.can_connect():
             self.unit.status = WaitingStatus("Waiting for bessd container to be ready")
-            event.defer()
```
ghislainbourgeois, line 220: Unfortunately, I do not think this will work because only `ConfigChanged` points to this handler, both `PebbleReady` point to their own handlers. This will need more refactoring.

ghislainbourgeois: I think we could complety remove both checks for `can_connect()` in this handler.

src/charm.py
```diff
@@ -230,6 +228,10 @@ def _on_bessd_pebble_ready(self, event: EventBase) -> None:
         """Handle Pebble ready event."""
         if not self.unit.is_leader():
             return
+        if not self._bessd_container.can_connect():
+            self.unit.status = WaitingStatus("Waiting for bessd container to be ready")
+            event.defer()
```
ghislainbourgeois, line 233: Here we can remove this `defer()`, this handler will be called again by `PebbleReady`.

src/charm.py
```diff
@@ -245,6 +247,10 @@ def _on_pfcp_agent_pebble_ready(self, event: EventBase) -> None:
         """Handle pfcp agent Pebble ready event."""
         if not self.unit.is_leader():
             return
+        if not self._pfcp_agent_container.can_connect():
+            self.unit.status = WaitingStatus("Waiting for pfcp agent container to be ready")
+            event.defer()
```
ghislainbourgeois, line 252: Here we can remove the `defer()`, this handler will be called again by `PebbleReady`.

ghislainbourgeois (changes requested):

src/charm.py
saltiyazan: Yes, I agree, I was thinking of handling both pebble ready events using this handler, but we would be in the same situation again of running the configuration script multiple times in a row quickly (For example `bessd_pebble_ready` then `pfcp_pebble_ready`)
However I found that we are checking if the config file has changed or not in and based on that we rewrite the config fiel `upf.json` or not, and we also restart the container or not based on that condition.
Wouldn't it be a better solution if we keep everything as is and we only call `_run_bess_configuration` which will run the `up4` script on the same condition as the restart for example, that being if the config file actually changed we run it, otherwise we don't.
so `self._run_bess_configuration()` would be inside the `if restart:`.

saltiyazan:
Issue addressed in https://github.com/canonical/sdcore-upf-operator/pull/26
closing.

PR #31, opened 2023-09-04
UPF support custom MTU sizes

gatici:
# Description

Adding optional configuration to set custom MTU sizes for UPF access and core interfaces.

# Checklist:


src/charm.py
```diff
@@ -571,6 +542,100 @@ def _gnb_subnet_config_is_valid(self) -> bool:
...
+                "addresses": [
+                    {
+                        "address": self._get_core_network_ip_config(),
+                    }
+                ],
+            },
+            "capabilities": {"mac": True},
+        }
+
+    def _get_core_interface_mtu_config(self) -> Optional[str]:
```
Gmerold, line 579: How about returning MTU as `int` right away? As `string` it's useless anyway.

src/charm.py
```diff
@@ -571,6 +542,100 @@ def _gnb_subnet_config_is_valid(self) -> bool:
...
+        return self.model.config.get("core-interface-mtu-size")
+
+    def _get_access_interface_mtu_config(self) -> Optional[str]:
```
Gmerold, line 582: As above.

src/charm.py
```diff
@@ -571,6 +542,100 @@ def _gnb_subnet_config_is_valid(self) -> bool:
...
+
+        Returns:
+            bool: Whether core interface MTU size is valid
+        """
+        if core_interface_mtu_size := self._get_core_interface_mtu_config() is None:
+            return True
+        return self._mtu_size_is_valid(int(core_interface_mtu_size))
+
+    @staticmethod
+    def _mtu_size_is_valid(mtu_size: Any) -> bool:
```
Gmerold, line 606: How about enforcing MTU being `int` on the signature level?

src/charm.py
```diff
@@ -72,22 +72,7 @@ def __init__(self, *args):
...
-                    interface=ACCESS_INTERFACE_NAME,
-                ),
-                NetworkAnnotation(
-                    name=CORE_NETWORK_ATTACHMENT_DEFINITION_NAME,
-                    interface=CORE_INTERFACE_NAME,
-                ),
-            ],
-            network_attachment_definitions_func=self._network_attachment_definitions_from_config,
-        )
+        self._kubernetes_multus = self._get_kubernetes_multus()
```
Gmerold, line 75: Probably an individual thing, but I liked the previous version more :) It would give an instant view on the NADs created by this charm.

src/charm.py
```diff
@@ -571,6 +542,100 @@ def _gnb_subnet_config_is_valid(self) -> bool:
...
+        """Checks whether the given MTU size is valid.
+
+        Returns:
+            bool: Whether the MTU size is valid
+        """
+        if not isinstance(mtu_size, int):
+            return False
+        return 0 < mtu_size <= 9000
+
+    def _get_kubernetes_multus(self) -> KubernetesMultusCharmLib:
```
Gmerold, line 616: I'm not quite sure about the method name and the docstring here. We're not really getting any Multus object , but an instance of a library. If we decide to keep this method (please see the related comment in `__init__`), perhaps something like `initialize_multus_library` would be more precise?

src/charm.py
```diff
@@ -224,6 +189,8 @@ def _on_config_changed(self, event: EventBase):
             self.unit.status = WaitingStatus("Waiting for pfcp agent container to be ready")
             event.defer()
             return
+        del self._kubernetes_multus
+        self._kubernetes_multus = self._get_kubernetes_multus()
```
Gmerold, line 193: Why do we need this? 
We'll get new instance of the charm every time Juju event is fired and we initialize the lib in `__init__` anyway.

src/charm.py
```diff
@@ -120,44 +105,24 @@ def _update_fiveg_n3_relation_data(self) -> None:
...
-                    {
-                        "address": self._get_core_network_ip_config(),
-                    }
-                ],
-            },
-            "capabilities": {"mac": True},
-        }
+
+        if (access_interface_mtu := self._get_access_interface_mtu_config()) is not None:
+            access_nad_config.update({"mtu": int(access_interface_mtu)})
```
Gmerold, line 115: This function will be called by the constructor, so I think we should validate the MTU before updating the NAD config.

src/charm.py
```diff
@@ -120,44 +105,24 @@ def _update_fiveg_n3_relation_data(self) -> None:
...
+
+        core_nad_config = self._get_core_nad_config()
         if (core_interface := self._get_core_interface_config()) is not None:
             core_nad_config.update({"type": "macvlan", "master": core_interface})
         else:
             core_nad_config.update({"type": "bridge", "bridge": "core-br"})
+
+        if (core_interface_mtu := self._get_core_interface_mtu_config()) is not None:
+            core_nad_config.update({"mtu": int(core_interface_mtu)})
```
Gmerold, line 124: As above.

Gmerold (changes requested):

src/charm.py

src/charm.py

src/charm.py
gatici: This static method is removed.

src/charm.py
gatici: I got back that change.

src/charm.py
gatici: I got back that change.

src/charm.py
gatici: We have strange issue, if we configure the MTU as default in config.yaml, it's set as expected. If we use `juju config upf  access-interface-mtu-size=1200`, it has no effect. According to juju debug-log, it deletes and creates network attachments but  access-br, core-br MTU sizes are stay same. 
```
unit-upf-0: 13:43:29 INFO unit.upf/0.juju-log HTTP Request: GET https://10.152.183.1/apis/k8s.cni.cncf.io/v1/namespaces/core/network-attachment-definitions "HTTP/1.1 200 OK"
unit-upf-0: 13:43:29 INFO unit.upf/0.juju-log HTTP Request: DELETE https://10.152.183.1/apis/k8s.cni.cncf.io/v1/namespaces/core/network-attachment-definitions/core-net "HTTP/1.1 200 OK"
unit-upf-0: 13:43:29 INFO unit.upf/0.juju-log NetworkAttachmentDefinition core-net deleted
unit-upf-0: 13:43:29 INFO unit.upf/0.juju-log HTTP Request: POST https://10.152.183.1/apis/k8s.cni.cncf.io/v1/namespaces/core/network-attachment-definitions "HTTP/1.1 201 Created"
unit-upf-0: 13:43:29 INFO unit.upf/0.juju-log NetworkAttachmentDefinition core-net created
unit-upf-0: 13:43:29 INFO unit.upf/0.juju-log HTTP Request: GET https://10.152.183.1/apis/apps/v1/namespaces/core/statefulsets/upf "HTTP/1.1 200 OK"
unit-upf-0: 13:43:29 INFO unit.upf/0.juju-log HTTP Request: GET https://10.152.183.1/apis/k8s.cni.cncf.io/v1/namespaces/core/network-attachment-definitions/access-net "HTTP/1.1 200 OK"
unit-upf-0: 13:43:29 INFO unit.upf/0.juju-log HTTP Request: GET https://10.152.183.1/apis/k8s.cni.cncf.io/v1/namespaces/core/network-attachment-definitions/core-net "HTTP/1.1 200 OK"
unit-upf-0: 13:43:29 INFO unit.upf/0.juju-log HTTP Request: GET https://10.152.183.1/apis/apps/v1/namespaces/core/statefulsets/upf "HTTP/1.1 200 OK"
unit-upf-0: 13:43:29 INFO unit.upf/0.juju-log HTTP Request: GET https://10.152.183.1/api/v1/namespaces/core/pods/upf-0 "HTTP/1.1 200 OK"
unit-upf-0: 13:43:29 INFO unit.upf/0.juju-log Default core network route created
unit-upf-0: 13:43:30 INFO unit.upf/0.juju-log No `fiveg_n3` relations found.
unit-upf-0: 13:43:30 INFO juju.worker.uniter.operation ran "config-changed" hook (via hook dispatching script: dispatch)
```

Do you want to check this issue together ?

src/charm.py

src/charm.py

config.yaml
```diff
@@ -18,6 +18,9 @@ options:
     type: string
     default: 192.168.250.1
     description: Gateway IP address to the Core Network.
+  core-interface-mtu-size:
+    type: int
+    description: MTU size of UPF's Core Interface which is allowed between 0 < MTU <= 9000.
```
gruyaume, line 23: Since we don't have a default value here in the config.yaml, we should tell users what will happen if they don't enter anything.

gruyaume (changes requested):

config.yaml
gruyaume: And should we have a more reasonable minimum than 0? Do we envision any use case that would need anyting under 1400?

lib/charms/kubernetes_charm_libraries/v0/multus.py
```diff
@@ -530,6 +530,8 @@ def _configure_network_attachment_definitions(self):
         2. Goes through the list of NetworkAttachmentDefinitions to create and create them all
         """
         network_attachment_definitions_to_create = self.network_attachment_definitions_func()
+        if not network_attachment_definitions_to_create:
```
gruyaume, line 533: We should not make this change here, here we should only `fetch-lib`. The actual maintainer of this lib is [this project](https://github.com/canonical/kubernetes-charm-libraries/blob/main/lib/charms/kubernetes_charm_libraries/v0/multus.py#L105) whose lib version in `5`.

config.yaml
gruyaume: We should also specify the unit (bytes)

gruyaume (changes requested):

config.yaml
markbeierl: vxlan underlay causes main ifaces to be 1442, so adding multus calico ifaces over that results in MTU being 1372.  I'd say the lowest should be around 1200 to be safe.

There is also work being done into "super" jumbo frames, so 9000 isn't even a hard upper limit yet: https://en.wikipedia.org/wiki/Jumbo_frame#Super_jumbo_frames

I would suggest upper limit, if we even want one, should be at leat 64k

src/charm.py
```diff
@@ -120,44 +120,32 @@ def _update_fiveg_n3_relation_data(self) -> None:
...
+        if invalid_configs := self._get_invalid_configs():
+            self.unit.status = BlockedStatus(
+                f"The following configurations are not valid: {invalid_configs}"
+            )
+            return None  # type: ignore
+
+        access_nad_config = self._get_access_nad_config()
+
+        if self._get_access_interface_mtu_config() is not None:
+            access_nad_config.update({"mtu": self._get_access_interface_mtu_config()})  # type: ignore[arg-type]  # noqa: E501
```
Gmerold, line 173: How about we move this to `_get_access_nad_config` just to keep the config consistent at all times?

src/charm.py
```diff
@@ -350,6 +338,10 @@ def _get_invalid_configs(self) -> list[str]:
             invalid_configs.append("core-gateway-ip")
         if not self._gnb_subnet_config_is_valid():
             invalid_configs.append("gnb-subnet")
+        if self._access_interface_mtu_size_is_valid() is False:
```
Gmerold, line 384:
```suggestion
        if not self._access_interface_mtu_size_is_valid():
```

src/charm.py
```diff
@@ -350,6 +338,10 @@ def _get_invalid_configs(self) -> list[str]:
...
+            invalid_configs.append("access-interface-mtu-size")
+        if self._core_interface_mtu_size_is_valid() is False:
```
Gmerold, line 386:
```suggestion
        if not self._core_interface_mtu_size_is_valid():
```

src/charm.py
```diff
@@ -571,6 +563,72 @@ def _gnb_subnet_config_is_valid(self) -> bool:
...
+                    {
+                        "address": self._get_core_network_ip_config(),
+                    }
+                ],
+            },
+            "capabilities": {"mac": True},
+        }
+
+    def _get_core_interface_mtu_config(self) -> Optional[int]:
+        return self.model.config.get("core-interface-mtu-size")  # type: ignore
```
Gmerold, line 696: Does this actually return `int`? I thought Juju configs are strings by default, aren't they?

src/charm.py
```diff
@@ -571,6 +563,72 @@ def _gnb_subnet_config_is_valid(self) -> bool:
...
+
+    def _get_access_interface_mtu_config(self) -> Optional[int]:
+        return self.model.config.get("access-interface-mtu-size")  # type: ignore
```
Gmerold, line 699: As above.

src/charm.py
```diff
@@ -571,6 +563,72 @@ def _gnb_subnet_config_is_valid(self) -> bool:
...
+
+    def _access_interface_mtu_size_is_valid(self) -> bool:
+        """Checks whether the access interface MTU size is valid.
+
+        Returns:
+            bool: Whether access interface MTU size is valid
+        """
+        if self._get_access_interface_mtu_config() is None:
```
Gmerold, line 707:
```suggestion
        if not self._get_access_interface_mtu_config():
```

src/charm.py
```diff
@@ -571,6 +563,72 @@ def _gnb_subnet_config_is_valid(self) -> bool:
...
+            return True
+        try:
+            return True if self._get_access_interface_mtu_config() in range(1, 9001) else False  # type: ignore[arg-type]  # noqa: E501
```
Gmerold, line 710:
```suggestion
            return self._get_access_interface_mtu_config() in range(1, 9001)
```

src/charm.py
```diff
@@ -571,6 +563,72 @@ def _gnb_subnet_config_is_valid(self) -> bool:
...
+        except ValueError:
+            return False
+
+    def _core_interface_mtu_size_is_valid(self) -> bool:
+        """Checks whether the core interface MTU size is valid.
+
+        Returns:
+            bool: Whether core interface MTU size is valid
+        """
+        if self._get_core_interface_mtu_config() is None:
```
Gmerold, line 720: As above.

src/charm.py
```diff
@@ -571,6 +563,72 @@ def _gnb_subnet_config_is_valid(self) -> bool:
...
+            return True
+        try:
+            return True if self._get_core_interface_mtu_config() in range(1, 9001) else False  # type: ignore[arg-type]  # noqa: E501
```
Gmerold, line 723: As above.

lib/charms/kubernetes_charm_libraries/v0/multus.py
```diff
@@ -148,6 +148,9 @@ def __init__(self, namespace: str):
         self.client = Client()
         self.namespace = namespace
 
+    def delete_pod(self, pod: str) -> None:
```
markbeierl, line 151: I don't think this belongs in the multus library.

config.yaml
markbeierl: Thinking more about this: other than it being an int, I don't think we should place any restriction on the value.  People won't be setting it if they don't know what they are doing, and there will always be an exception to whatever rule we come up with

lib/charms/kubernetes_charm_libraries/v0/multus.py
gatici: Yes, it does not belong here. This is draft. I will push another version.

src/charm.py
gatici: Yes, it is returning int.

gatici:
This is the change in the library: https://github.com/canonical/kubernetes-charm-libraries/pull/12

src/charm.py
markbeierl: What I am seeing is that even though the NADs are deleted, the ifaces remain on the host until reboot.  I suspect there is something missing in the cleanup

gatici: I saw the same behaviour, although there is no NAD, bridges in the host machines stays there. This is also happened by using kubectl with some sample pods.

config.yaml
gatici: Done

lib/charms/kubernetes_charm_libraries/v0/multus.py
gatici: Done

src/charm.py

src/charm.py

src/charm.py
gatici: There is no issue to set it as int.  I do not see any problem here.

src/charm.py
gatici: We need to use  "if self._get_access_interface_mtu_config() is None". Reason is use may set 0 as value (which is invalid) and following code block return True which should be False. 
 if not self._get_access_interface_mtu_config():
      return True
      
 If config is empty, the following block is fine.
 if self._get_access_interface_mtu_config() is None:
     return True

src/charm.py

src/charm.py

lib/charms/kubernetes_charm_libraries/v0/multus.py
gatici: Done

gatici:
Waiting for PR: https://github.com/canonical/kubernetes-charm-libraries/pull/12 to be merged.

lib/charms/kubernetes_charm_libraries/v0/multus.py
```diff
@@ -619,3 +627,15 @@ def _on_remove(self, event: RemoveEvent) -> None:
                 self.kubernetes.delete_network_attachment_definition(
                     name=network_attachment_definition.metadata.name  # type: ignore[union-attr]
                 )
+
+    def delete_pod(self) -> None:
+        """Delete the pod."""
+        self.kubernetes.delete_pod(self._pod)
```
markbeierl, line 633: Would it be ok to just do
`self.kubernetes.client.delete(Pod, pod_name, namespace)`
?

markbeierl (changes requested):

src/charm.py
```diff
@@ -225,6 +246,13 @@ def _on_config_changed(self, event: EventBase):
             event.defer()
             return
         self._on_bessd_pebble_ready(event)
+        if self._restart_pod:
+            try:
+                self._kubernetes_multus.delete_pod()
```
markbeierl, line 251: Would it be ok to just do
`self._kubernetes_multus.client.delete(Pod, pod_name, namespace)`
here instead of modifying the lib?

lib/charms/kubernetes_charm_libraries/v0/multus.py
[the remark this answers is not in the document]
gatici: This is already done: https://github.com/canonical/kubernetes-charm-libraries/pull/12. Waiting in review. We also need update this file to make it compatible with the new library version.

src/charm.py
gatici: I think we require to reuse the delete method also in other charms which uses multus. Adding it to the library seemed reasonable to me as it is easy to manage.

config.yaml
```diff
@@ -18,6 +18,13 @@ options:
     type: string
     default: 192.168.250.1
     description: Gateway IP address to the Core Network.
+  core-interface-mtu-size:
+    type: int
+    description: |
+      MTU size of UPF's Core Interface which is allowed
+      between 1200 < MTU <= 65536 (bytes).
+      If config is not provided, Multus creates interface
+      with default MTU size which is 1500 bytes.
```
patriciareinoso, line 27: I am a little confused. I don't see this part being implemented in the code. Is it done by Multus so it is out of the scope of this charm ?
should we set a default value = 1500 to the core-inteface-mtu-size config instead of handling the <None> case in the config? 

Same comment for access-inteface-mtu-size

src/charm.py
```diff
@@ -95,6 +98,38 @@ def __init__(self, *args):
...
+    def _is_pod_restart_required(self) -> None:
+        """Checks if existing NAD config is different then the required ones which are set in the configuration  # noqa: E501, W505.
+
+        If there is difference, sets the _restart_pod attribute to True.
+        """
+        if (
+            self.get_nad_active_mtu_size(ACCESS_NETWORK_ATTACHMENT_DEFINITION_NAME)
+            != self._get_access_interface_mtu_config()  # noqa: W503
+        ):
+            self._restart_pod = True
```
patriciareinoso, line 110: we can return after this. no need to check the second condition

config.yaml
gatici: Default MTU size is set by Multus. We do not need to do something to set default MTU. We are explaining here that if user does not configure it, interface comes with MTU size 1500.

src/charm.py
gatici: Good catch. We can add a return here.

src/charm.py
patriciareinoso, line 251: I don't understand much the order on the execution flow. But why basing the decision of restarting the pod over a flag set on _network_attachment_definitions_from_config method instead of detecting the config change here are deleting the pod right away?

src/charm.py
gatici: Done

src/charm.py
```diff
@@ -571,6 +603,104 @@ def _gnb_subnet_config_is_valid(self) -> bool:
...
+                "addresses": [
+                    {
+                        "address": self._get_access_network_ip_config(),
+                    }
+                ],
+            },
+            "capabilities": {"mac": True},
+        }
+        if self._get_access_interface_mtu_config():
+            config.update({"mtu": self._get_access_interface_mtu_config()})  # type: ignore
```
patriciareinoso, line 634: I understand that  the _get_access_nad_config method is not executed if the access interface mtu config is invalid. I am just worried that someone in the future wants to reuse this methods and find themselves with an invalid valid of the access interface mtu config because we don't check whether the value is valid

src/charm.py
```diff
@@ -95,6 +98,38 @@ def __init__(self, *args):
             self.fiveg_n3_provider.on.fiveg_n3_request, self._on_fiveg_n3_request
         )
 
+    def _is_pod_restart_required(self) -> None:
+        """Checks if existing NAD config is different then the required ones which are set in the configuration  # noqa: E501, W505.
```
patriciareinoso, line 102:
```suggestion
        """Checks if existing NAD config is different than the required ones which are set in the configuration  # noqa: E501, W505.
```

src/charm.py
```diff
@@ -95,6 +98,38 @@ def __init__(self, *args):
...
+            != self._get_access_interface_mtu_config()  # noqa: W503
+        ):
+            self._restart_pod = True
+        if (
+            self.get_nad_active_mtu_size(CORE_NETWORK_ATTACHMENT_DEFINITION_NAME)
+            != self._get_core_interface_mtu_config()  # noqa: W503
+        ):
+            self._restart_pod = True
+
+    def get_nad_active_mtu_size(self, nad_name: str) -> Optional[int]:
```
patriciareinoso, line 119:
```suggestion
    def _get_nad_active_mtu_size(self, nad_name: str) -> Optional[int]:
```

src/charm.py
gatici: KubernetesMultusCharmLib library run _configure_multus method which deletes the old NAD definitions and creates new NAD definitions with latest config settings when config_changed event is fired. In config_changed method of charm if we try to detect if pod restart is required or not, it returns False. So this check needs to be done before config_changed run and restart needs to be done after new NADs are created to make the new NADs take effect.

```
self.framework.observe(charm.on.config_changed, self._configure_multus)
```

tests/unit/test_charm.py
```diff
@@ -9,9 +9,16 @@
...
+from sample_nad_defitions import modified_nad_descriptions, original_nad_descriptions
 
 from charm import UPFOperatorCharm
 
+MULTUS_LIBRARY_PATH = "charms.kubernetes_charm_libraries.v0.multus"
+TOO_BIG_MTU_SIZE = 80000
+TOO_SMALL_MTU_SIZE = 0
+VALID_MTU_SIZE_1 = 9000
+VALID_MTU_SIZE_2 = 1800
+
```
patriciareinoso, line 22: It would be worth to test edge cases : 0, 1200, 1201, 65636, 65637, ...

gatici: OK, I will expand the tests to include them.

src/charm.py
gatici: OK. Done

src/charm.py
gatici: Ok, Done

src/charm.py
gatici: We are not validating the config settings in _get_acccess_nad_config method. But, this is valid for all config settings used in this method (gnb-subnet, access-interface-gw, etc) and not only for access-interface-mtu.  I found it useful to validate any config before constituting the NAD configs.  If any config is invalid, the execution should not come to NAD creation. Validating manual user inputs as much as earlier seems more beneficial. I think it will not be a problem to reuse these methods and developer could easily discover it.

tests/unit/test_charm.py
gatici: Done

src/charm.py
```diff
@@ -95,6 +98,40 @@ def __init__(self, *args):
...
+    def _is_pod_restart_required(self) -> None:
+        """Checks if existing NAD config is different from the required ones which are set in the configuration  # noqa: E501, W505.
+
+        If there is difference, sets the _restart_pod attribute to True.
+        """
+        if (
+            self._get_nad_active_mtu_size(ACCESS_NETWORK_ATTACHMENT_DEFINITION_NAME)
+            != self._get_access_interface_mtu_config()  # noqa: W503
+        ):
+            self._restart_pod = True
```
gruyaume, line 110: I'm not a huge fan of setting attributes here. The problem with charms is that it's easy to get confused. It's possible to set the value on 1 event and this value is not accessible on follow up events. I rather us call this method directly every time we need it.

In other words, let's have this method return a bool instead of setting "restart_pod".

src/charm.py
```diff
@@ -119,45 +156,31 @@ def _update_fiveg_n3_relation_data(self) -> None:
...
-                    }
-                ],
-            },
-            "capabilities": {"mac": True},
-        }
-        if (access_interface := self._get_access_interface_config()) is not None:
+        """Returns list of Multus NetworkAttachmentDefinitions to be created based on config.
+
+        Checks if the config settings are correct before creating any NAD.
+        It also detects the MTU size change in NAD definitions to trigger the restart of pod.
```
gruyaume, line 162: Same comment around "nad definition"

gruyaume (changes requested):

src/charm.py
gatici: Done

src/charm.py
gatici: Done

src/charm.py
```diff
@@ -95,6 +98,40 @@ def __init__(self, *args):
...
+
+    def _get_nad_active_mtu_size(self, nad_name: str) -> Optional[int]:
+        """Gets active MTU size for a given NAD name.
+
+        Args:
+            nad_name    (str):  NAD name
+
+        Returns:
+            mtu_size    (int/None):  Active MTU size for a given NAD
+                                If NAD definition is not found, returns None
```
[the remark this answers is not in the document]
gatici: Done

tests/unit/sample_nad_defitions.py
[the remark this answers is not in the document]
gatici: Done

src/charm.py
markbeierl: It looks like manual intervention is required: in order to change the MTU, all applications using the same network must be removed.  Next remove the bridge using `ip link down core-br` or whatever the name is.  You can then redeploy the app with a new MTU and it will take effect

src/charm.py
```diff
@@ -53,6 +55,7 @@ def __init__(self, *args):
         super().__init__(*args)
         self._bessd_container_name = self._bessd_service_name = "bessd"
         self._routectl_service_name = "routectl"
+        self._restart_pod = False
```
gruyaume, line 58: Similar to my other comment, let's not have/set a global value for restart pod. Hooks that need to know this should call `_is_pod_restart_required` on their own instead of looking at this value. We shouldn't need `self._restart_pod` anywhere here.

gruyaume (changes requested):

src/charm.py
gatici: There is a reason to do store the boolean in a variable before running it under config changed after new NADs are created. 
KubernetesMultusCharmLib library run _configure_multus method which deletes the old NAD definitions and creates new NAD definitions with latest config settings when config_changed event is fired. In config_changed method of charm if we try to detect whether pod restart is required or not, it always returns False. So this check needs to be done before config_changed run and restart needs to be done after new NADs are created to make the new NADs take effect.

gatici: Pod restart  needs to wait till new NADs are created and pod restart detection should be done before old NADs deleted.  I was simply storing the status to take the action later. I am open to any other suggestions to overcome this situation.

src/charm.py
```diff
@@ -144,45 +162,30 @@ def _update_fiveg_n3_relation_data(self) -> None:
...
+        """Returns list of Multus NetworkAttachmentDefinitions to be created based on config.
+
+        Checks if the config settings are correct before creating any NAD.
+        If the NAD config is invalid, returns None as a protection not to process invalid configs.
+        """
+        if invalid_configs := self._get_invalid_configs():
+            self.unit.status = BlockedStatus(
+                f"The following configurations are not valid: {invalid_configs}"
+            )
+            return None  # type: ignore
```
ghislainbourgeois, line 174: This should not return `None`, but an empty list according to the type hint. 

I understand however that this might be necessary in order to not remove all `NADs` if the config is updated to an invalid state. In that case, the type hints should be updated here and in the lib.

src/charm.py
ghislainbourgeois: It does not return `int`, but `Optional[str]`.

Please remove the `# type: ignore` and update the type hint for the return type of the method.

src/charm.py
ghislainbourgeois: As above.

src/charm.py
```diff
@@ -622,6 +629,104 @@ def _get_cpu_extensions() -> list[str]:
...
+                "addresses": [
+                    {
+                        "address": self._get_core_network_ip_config(),
+                    }
+                ],
+            },
+            "capabilities": {"mac": True},
+        }
+        if self._get_core_interface_mtu_config():
+            config.update({"mtu": self._get_core_interface_mtu_config()})  # type: ignore
```
ghislainbourgeois, line 681:
```suggestion
        if core_mtu := self._get_core_interface_mtu_config():
            config.update({"mtu": core_mtu})
```

In addition to fixing the type hint for the `_get_core_interface_mtu_config()` method, this should work.

src/charm.py
```diff
@@ -622,6 +629,104 @@ def _get_cpu_extensions() -> list[str]:
...
+                "addresses": [
+                    {
+                        "address": self._get_access_network_ip_config(),
+                    }
+                ],
+            },
+            "capabilities": {"mac": True},
+        }
+        if self._get_access_interface_mtu_config():
+            config.update({"mtu": self._get_access_interface_mtu_config()})  # type: ignore
```
ghislainbourgeois, line 658:
```suggestion
        if access_mtu := self._get_access_interface_mtu_config():
            config.update({"mtu": access_mtu})
```

src/charm.py
```diff
@@ -622,6 +629,104 @@ def _get_cpu_extensions() -> list[str]:
...
+
+        Returns:
+            bool: Whether access interface MTU size is valid
+        """
+        if self._get_access_interface_mtu_config() is None:
+            return True
+        try:
+            return self._get_access_interface_mtu_config() in range(1200, 65537)
+        except ValueError:
+            return False
```
ghislainbourgeois, line 715: This is more efficient:

```suggestion
        return (1200 <= self._get_access_interface_mtu_config() < 65537)
```

ghislainbourgeois (changes requested):

src/charm.py
gatici: I will update the type hints.

src/charm.py
gatici: def _get_core_interface_mtu_config(self) -> Optional[int]:
        return self.model.config.get("core-interface-mtu-size")  

I tested the return value before writing Optional[int]. If we set int, here the return type was int. Anyway in the config.yaml I set the value type as int and it does not allow me to set a string.

gatici: You are suggesting to set Optional[str] to fix type for below code block.  If we will have type problem in the following line which is type str could not be compared in that way.
 (1200 <= self._get_access_interface_mtu_config() < 65537)
 ```
 def _get_access_interface_mtu_config(self) -> Optional[int]:
        return self.model.config.get("access-interface-mtu-size") 
 ```

src/charm.py
gatici: OK

src/charm.py
gatici: OK

src/charm.py
gatici: It does not seems correct to me. We are indicating type int in config.yaml, however setting it Optional[str] just to fix the type problem here but creates another typing issue in the validation of value.

src/charm.py
gatici: Done

src/charm.py
ghislainbourgeois: I agree that we will need to do an explicit `int` conversion, but it happened anyway implicitly before.

src/charm.py
```diff
@@ -622,6 +634,104 @@ def _get_cpu_extensions() -> list[str]:
...
+
+    def _get_core_interface_mtu_config(self) -> Optional[Any]:
+        """Get Core interface MTU size.
+
+        Returns:
+            mtu_size (int/None): If MTU size is not configured return None
+                                    If it is set, returns the configured value expected in int
+
+        """
+        return self.model.config.get("core-interface-mtu-size")
```
ghislainbourgeois, line 697: If we want this to return an `int`, we can explicitly do the conversion here and set the return type:

```suggestion
    def _get_core_interface_mtu_config(self) -> Optional[int]:
        """Get Core interface MTU size.

        Returns:
            mtu_size (int/None): If MTU size is not configured return None
                                    If it is set, returns the configured value expected in int

        """
        if mtu_size := self.model.config.get("core-interface-mtu-size"):
            return int(mtu_size)
        return None
```

Otherwise, we should set the typing explicitly to `Optional[str]` and do the explicit conversion when we need it as an `int`. 

I think I prefer a conversion to `int` only when we do the validation.

src/charm.py
```diff
@@ -622,6 +634,104 @@ def _get_cpu_extensions() -> list[str]:
...
+
+    def _get_access_interface_mtu_config(self) -> Optional[Any]:
+        """Get Access interface MTU size.
+
+        Returns:
+            mtu_size (int/None): If MTU size is not configured return None
+                                    If it is set, returns the configured value expected in int
+
+        """
+        return self.model.config.get("access-interface-mtu-size")
```
ghislainbourgeois, line 707: As above

ghislainbourgeois (changes requested):

src/charm.py
gatici: Returned as str, then str -> int validation is done in validation.

src/charm.py
gatici: Returned as str, then str -> int validation is done in validation.

src/charm.py
gatici: Returned as str, then str -> int validation is done in validation.

src/charm.py
gatici: Returned as str, then str -> int validation is done in validation.

src/charm.py

src/charm.py
```diff
@@ -672,6 +668,104 @@ def _get_cpu_extensions() -> list[str]:
...
+    def _access_interface_mtu_size_is_valid(self) -> bool:
+        """Checks whether the access interface MTU size is valid.
+
+        Returns:
+            bool: Whether access interface MTU size is valid
+        """
+        if (access_mtu := self._get_access_interface_mtu_config()) is None:
+            return True
+        try:
+            return 1200 <= int(access_mtu) < 65537
```
ghislainbourgeois, line 752: The IPv4 header Total Length is a 16 bit field.

```suggestion
            return 1200 <= int(access_mtu) <= 65535
```

src/charm.py
```diff
@@ -672,6 +668,104 @@ def _get_cpu_extensions() -> list[str]:
...
+    def _core_interface_mtu_size_is_valid(self) -> bool:
+        """Checks whether the core interface MTU size is valid.
+
+        Returns:
+            bool: Whether core interface MTU size is valid
+        """
+        if (core_mtu := self._get_core_interface_mtu_config()) is None:
+            return True
+        try:
+            return 1200 <= int(core_mtu) < 65537
```
ghislainbourgeois, line 765:
```suggestion
            return 1200 <= int(core_mtu) <= 65535
```

tests/unit/test_charm.py
```diff
@@ -15,6 +15,13 @@
 
 from charm import IncompatibleCPUError, UPFOperatorCharm
 
+MULTUS_LIBRARY_PATH = "charms.kubernetes_charm_libraries.v0.multus"
+TOO_BIG_MTU_SIZE = 65537  # Out of range
+TOO_SMALL_MTU_SIZE = 1199  # Out of range
+ZERO_MTU_SIZE = 0  # Out of range
+VALID_MTU_SIZE_1 = 65536  # Upper edge value
```
ghislainbourgeois, line 22:
```suggestion
TOO_BIG_MTU_SIZE = 65536  # Out of range
TOO_SMALL_MTU_SIZE = 1199  # Out of range
ZERO_MTU_SIZE = 0  # Out of range
VALID_MTU_SIZE_1 = 65535  # Upper edge value
```

tests/unit/test_charm.py
```diff
@@ -574,3 +590,219 @@ def test_when_remove_then_external_service_is_deleted(self, patch_client):
...
+                "gnb-subnet": "192.168.251.0/24",
+                "core-ip": "192.168.250.3/24",
+                "core-gateway-ip": "192.168.250.1",
+                "core-interface-mtu-size": VALID_MTU_SIZE_1,
+            }
+        )
+        nads = self.harness.charm._network_attachment_definitions_from_config()
+        for nad in nads:
+            config = json.loads(nad.spec["config"])
+            self.assertEqual(config["mtu"], 65536)
```
ghislainbourgeois, line 630:
```suggestion
            self.assertEqual(config["mtu"], VALID_MTU_SIZE_1)
```

ghislainbourgeois (changes requested):
Some small changes, otherwise LGTM

src/charm.py
gatici: Done

tests/unit/test_charm.py
gatici: Done

tests/unit/test_charm.py
gatici: Done

tests/unit/test_charm.py
```diff
@@ -730,3 +749,219 @@ def test_when_remove_then_external_service_is_deleted(self, patch_client):
...
+        )
+        self.assertEqual(
+            self.harness.model.unit.status,
+            BlockedStatus(
+                "The following configurations are not valid: ['access-interface-mtu-size', 'core-interface-mtu-size']"  # noqa: E501, W505
+            ),
+        )
+
+    @patch(
+        f"{MULTUS_LIBRARY_PATH}.KubernetesMultusCharmLib._configure_network_attachment_definitions",  # noqa: E501
```
gruyaume, line 813: Same comment as above

tests/unit/test_charm.py
```diff
@@ -730,3 +749,219 @@ def test_when_remove_then_external_service_is_deleted(self, patch_client):
...
+        patch_list_na_definitions.return_value = original_nads
+        self.harness.update_config(key_values={"core-interface-mtu-size": VALID_MTU_SIZE_1})
+        modified_nads = self.harness.charm._network_attachment_definitions_from_config()
+        patch_list_na_definitions.return_value = modified_nads
+        self.harness.update_config(key_values={"core-interface-mtu-size": VALID_MTU_SIZE_2})
+        self.assertEqual(patch_delete_pod.call_count, 2)
+
+    @patch(f"{MULTUS_LIBRARY_PATH}.KubernetesClient", new=Mock)
+    @patch(
+        f"{MULTUS_LIBRARY_PATH}.KubernetesMultusCharmLib._network_attachment_definition_created_by_charm"  # noqa: E501, W505
```
gruyaume, line 898: Same comment as above

tests/unit/test_charm.py
```diff
@@ -730,3 +749,219 @@ def test_when_remove_then_external_service_is_deleted(self, patch_client):
...
+                "core-interface-mtu-size": VALID_MTU_SIZE_1,
+            }
+        )
+        nads = self.harness.charm._network_attachment_definitions_from_config()
+        for nad in nads:
+            config = json.loads(nad.spec["config"])
+            self.assertEqual(config["mtu"], 65535)
+
+    @patch(
+        f"{MULTUS_LIBRARY_PATH}.KubernetesMultusCharmLib._configure_network_attachment_definitions",  # noqa: E501
```
gruyaume, line 792: We should not patch the lib private methods in our charm tests. There is no guarantee that this will remain the same. In general, we should only patch public method calls

src/charm.py
```diff
@@ -250,46 +252,31 @@ def _get_n4_upf_hostname(self) -> str:
...
-            },
-            "capabilities": {"mac": True},
-        }
-        if (access_interface := self._get_access_interface_config()) is not None:
+    def _network_attachment_definitions_from_config(
+        self,
+    ) -> Optional[list[NetworkAttachmentDefinition]]:
+        """Returns list of Multus NetworkAttachmentDefinitions to be created based on config.
+
+        Checks if the config settings are correct before creating any NAD.
```
gruyaume, line 260: It's not this method that does this

tests/unit/test_charm.py
```diff
@@ -730,3 +749,219 @@ def test_when_remove_then_external_service_is_deleted(self, patch_client):
...
+        patch_nad_created_by_charm,
+    ):
+        """Delete pod is called for the first config change, setting the same config value does not trigger pod restarts."""  # noqa: E501, W505
+        patch_exists.return_value = True
+        patch_multus_is_ready.return_value = True
+        self.harness.set_can_connect(container="bessd", val=True)
+        self.harness.set_can_connect(container="pfcp-agent", val=True)
+        self.harness.set_leader(is_leader=True)
+        patch_nad_created_by_charm.return_value = True
+        original_nads = self.harness.charm._network_attachment_definitions_from_config()
```
gruyaume, line 922: We should not patch our private methods.

In the tests we should have no reference to private methods except the top level event handlers (ex. `_on_install`). This applies at a couple of places.

gruyaume (changes requested):

src/charm.py
gatici: Done

tests/unit/test_charm.py
gatici: Done

tests/unit/test_charm.py
gatici: Done

tests/unit/test_charm.py
gatici: Done

tests/unit/test_charm.py
gatici: Done

src/charm.py
```diff
@@ -250,46 +252,29 @@ def _get_n4_upf_hostname(self) -> str:
...
-        }
-        if (access_interface := self._get_access_interface_config()) is not None:
+    def _network_attachment_definitions_from_config(
+        self,
+    ) -> Optional[list[NetworkAttachmentDefinition]]:
+        """Returns list of Multus NetworkAttachmentDefinitions to be created based on config.
+
+        Returns:
+            network_attachment_definitions: list[NetworkAttachmentDefinition] if config is valid
+                                            None if config is invalid
```
ghislainbourgeois, line 262: This does not seem to ever return None with the latest changes.
```suggestion
    def _network_attachment_definitions_from_config(
        self,
    ) -> list[NetworkAttachmentDefinition]:
        """Returns list of Multus NetworkAttachmentDefinitions to be created based on config.

        Returns:
            network_attachment_definitions: list[NetworkAttachmentDefinition]
```

ghislainbourgeois (changes requested):
Small nitpick, otherwise LGTM

src/charm.py

src/charm.py
```diff
@@ -250,46 +252,28 @@ def _get_n4_upf_hostname(self) -> str:
...
-                        "address": self._get_access_network_ip_config(),
-                    }
-                ],
-            },
-            "capabilities": {"mac": True},
-        }
-        if (access_interface := self._get_access_interface_config()) is not None:
+    def _network_attachment_definitions_from_config(
+        self,
+    ) -> Optional[list[NetworkAttachmentDefinition]]:
```
ghislainbourgeois, line 257:
```suggestion
    ) -> list[NetworkAttachmentDefinition]:
```
The `Optional` type hint was left behind

ghislainbourgeois (changes requested):

ghislainbourgeois (approved):

gruyaume (approved):

markbeierl (approved):

gatici:
In this PR, there are unsigned commits and having problem while signing some old commits. So https://github.com/canonical/sdcore-upf-operator/pull/44 is created as a replication of this PR with signing and squashing all the commits.

PR #36, opened 2023-09-07
Use LoadBalancer Service

DanielArndt:
# Description

Exposes an external LoadBalancer service for the UPF, similar to https://github.com/canonical/sdcore-amf-operator/pull/24 for the AMF

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -23,9 +23,13 @@ from charms.prometheus_k8s.v0.prometheus_scrape import (  # type: ignore[import]
 )
 from charms.sdcore_upf.v0.fiveg_n3 import N3Provides  # type: ignore[import]
 from jinja2 import Environment, FileSystemLoader
-from lightkube.models.core_v1 import ServicePort
+from lightkube.core.client import Client
+from lightkube.models.core_v1 import ServicePort, ServiceSpec
 from lightkube.models.meta_v1 import ObjectMeta
-from ops.charm import CharmBase, EventBase
+from lightkube.resources.core_v1 import Service
+from ops import RemoveEvent
+from ops.charm import CharmBase
+from ops.framework import EventBase
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, Container, ModelError, WaitingStatus
 from ops.pebble import ExecError, Layer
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -94,6 +98,54 @@ def __init__(self, *args):
...
+                },
+            ),
+            spec=ServiceSpec(
+                selector={
+                    "app.kubernetes.io/name": self.app.name,
+                },
+                ports=[
+                    ServicePort(name="pfcp", port=8805, protocol="UDP"),
+                ],
+                type="LoadBalancer",
```
gatici, line 126: This will create a service with a port assigned by Loadbalancer. 
service/upf-external                          LoadBalancer   10.152.183.224   10.0.0.2      8805:31523/UDP      2m
I have a question here.
Do we need to allow to define any specific port by user ? At the moment, it will be automatically assigned by Loadbalancer service. (31523)

Second issue:
I am not sure, if we need to consider this at the moment but for public cloud environments, we may need to provide Loadbalancer IP, if cloud-provider requires to specify the loadBalancerIP.

Gmerold (changes requested):
Two general comments:
1. I think we could mention in the README that it is possible to expose UPF, but LB is required for that. We could also have an integration test validating the exposure. WDYT?
2. Please add docstrings :)

src/charm.py
```diff
@@ -94,6 +98,54 @@ def __init__(self, *args):
...
+                labels={
+                    "app.kubernetes.io/name": self.app.name,
+                },
+            ),
+            spec=ServiceSpec(
+                selector={
+                    "app.kubernetes.io/name": self.app.name,
+                },
+                ports=[
+                    ServicePort(name="pfcp", port=8805, protocol="UDP"),
```
gatici, line 124: I suggest to create a variables for pfcp port name and value (8805).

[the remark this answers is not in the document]
DanielArndt: I believe the intention was to have both the internal and the external one. I'd have to rely on @ghislainbourgeois here to explain the benefit of doing it that way.

DanielArndt:
> Two general comments:
> 
> 1. I think we could mention in the README that it is possible to expose UPF, but LB is required for that. We could also have an integration test validating the exposure. WDYT?

Sure, I can add that to the README. 

I'd be happy to give integration tests a shot, but at this point I'm not quite sure what that would look like. I'll have to tackle those when I get back on Wednesday.

> 2. Please add docstrings :)

Happy to add docstrings where you think they would be useful. I'm not a fan of docstrings on internal functions where the docstring just reiterates the function name. i.e.

```
def _create_external_upf_service(self) -> None:
    """Creates the external UPF service."""
````

The above seems like a waste, but maybe there is something you'd like me to be more explicit about?

File: src/charm.py [hunk 3]
```diff
@@ -94,6 +98,54 @@ def __init__(self, *args):
         self.framework.observe(
             self.fiveg_n3_provider.on.fiveg_n3_request, self._on_fiveg_n3_request
         )
+        self.framework.observe(self.on.install, self._on_install)
+        self.framework.observe(self.on.remove, self._on_remove)
+
+    def _on_install(self, event: EventBase) -> None:
+        self._create_external_upf_service()
+
+    def _create_external_upf_service(self) -> None:
```
DanielArndt, line 107: Could be a staticmethod

src/charm.py
```diff
@@ -94,6 +98,54 @@ def __init__(self, *args):
...
+            ),
+        )
+
+        client.create(service)
+        logger.info("Created external UPF service")
+
+    def _on_remove(self, event: RemoveEvent) -> None:
+        self._delete_external_upf_service()
+
+    def _delete_external_upf_service(self) -> None:
```
DanielArndt, line 136: Could be a staticmethod

src/charm.py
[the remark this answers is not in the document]
Gmerold: I don't think we need both. The main goal of this story is to expose the pfcp service through the load balancer, to be able to deploy UPF separately from the rest of the 5G core.

src/charm.py
Gmerold: Ad. 1. I don't think it's a problem. What matters is that we need our service to be available on port `8805`. Which internal port it maps to doesn't really matter.
Ad. 2. I'm not sure I understand the problem. Would you mind giving an example?

src/charm.py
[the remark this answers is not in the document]
DanielArndt: More than happy to remove it, but @ghislainbourgeois was the one who suggested leaving it, so I just want to make sure he didn't have something else in mind. I'll try to grab his attention.

src/charm.py
DanielArndt: Oops, no it can't.

src/charm.py
Gmerold: OK, that explains it :) I didn't know that option ever existed :D

gatici:
We need to rebase it on top of main.

ghislainbourgeois (approved):

src/charm.py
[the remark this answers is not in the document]
Gmerold: Either way is fine for me. I really don't expect too many users of our tutorial at this point anyway ;) 
My only point here was that we probably don't want to keep both.
@DanielArndt I leave it to you to decide which approach you prefer.

Gmerold (approved):

gatici (approved):
Needs to be rebased before merging. LGTM.

src/charm.py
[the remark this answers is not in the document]
DanielArndt: I'd prefer the incremental approach, so I've made follow-up tasks for this.

PR #40, opened 2023-09-20
(Fix) Fixing relation data

Gmerold:
# Description

Fixes a bug in `fiveg_n3` and `fiveg_n4` libs. 
Bug was caused by the fact that the provider side would set the app relation data, but requirer would pull the unit relation data.

# Checklist:

### DIFF
--- a/lib/charms/sdcore_upf/v0/fiveg_n3.py
+++ b/lib/charms/sdcore_upf/v0/fiveg_n3.py
@@ -93,7 +93,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
 
 PYDEPS = ["pydantic", "pytest-interface-tester"]
 
@@ -260,6 +260,6 @@ class N3Requires(Object):
             event (RelationChangedEvent): Juju event
         """
         relation_data = event.relation.data
-        upf_ip_address = relation_data[event.unit].get("upf_ip_address")  # type: ignore[index]
+        upf_ip_address = relation_data[event.app].get("upf_ip_address")  # type: ignore[index]
         if upf_ip_address:
             self.on.fiveg_n3_available.emit(upf_ip_address=upf_ip_address)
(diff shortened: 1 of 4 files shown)


lib/charms/sdcore_upf/v0/fiveg_n3.py
```diff
@@ -260,6 +260,6 @@ def _on_relation_changed(self, event: RelationChangedEvent) -> None:
             event (RelationChangedEvent): Juju event
         """
         relation_data = event.relation.data
-        upf_ip_address = relation_data[event.unit].get("upf_ip_address")  # type: ignore[index]
+        upf_ip_address = relation_data[event.app].get("upf_ip_address")  # type: ignore[index]
```
saltiyazan, line 263: shouldn't we check if unit is leader?

gatici: I think, we do not need this check here. Because non-leaders can read from other application databags (event.relation.data[event.app]).  And those are all read actions.

tests/unit/lib/charms/sdcore_upf/v0/test_fiveg_n3_requirer.py
```diff
@@ -27,7 +27,7 @@ def test_given_relation_with_n3_profider_when_fiveg_n3_available_event_then_n3_i
 
         self.harness.update_relation_data(
             relation_id=relation_id,
-            app_or_unit="whatever-app/0",
+            app_or_unit="whatever-app",
```
gatici, line 30: The test name has a typo:
profider -> provider
test_given_relation_with_n3_provider_when_fiveg_n3_available_event_then_n3_information_is_provided

gatici (approved):
LGTM

lib/charms/sdcore_upf/v0/fiveg_n3.py
Gmerold: As discussed in the chat, I don't think it's needed here as this is a requirer side. I don't think we care who reads the data. We could be checking leadership on the provider side, but then:
1. Should this be handled by the charm or lib?
2. Do we need it now if we don't support scaling?

saltiyazan (approved):

tests/unit/lib/charms/sdcore_upf/v0/test_fiveg_n3_requirer.py
Gmerold: Good catch :) 
Fixed.

PR #41, opened 2023-09-20
Use custom event to trigger multus configuration

gatici:
# Description

This PR updates the multus library to include the fix for https://github.com/canonical/kubernetes-charm-libraries/issues/16 and implements the necessary changes in the charm part.

# Checklist:

### DIFF
--- a/lib/charms/kubernetes_charm_libraries/v0/multus.py
+++ b/lib/charms/kubernetes_charm_libraries/v0/multus.py
@@ -3,7 +3,8 @@
 
 """Charm Library used to leverage the Multus Kubernetes CNI in charms.
 
-- On config-changed, it will:
+- On a BoundEvent (e.g. self.on.nad_config_changed which is originated from NadConfigChangedEvent),
+ it will:
   - Configure the requested network attachment definitions
   - Patch the statefulset with the necessary annotations for the container to have interfaces
     that use those new network attachments.
@@ -20,7 +21,20 @@ from charms.kubernetes_charm_libraries.v0.multus import (
     NetworkAnnotation
 )
 
+class NadConfigChangedEvent(EventBase):
+
+    def __init__(self, handle: Handle):
+        super().__init__(handle)
+
+
+class KubernetesMultusCharmEvents(CharmEvents):
+
+    nad_config_changed = EventSource(NadConfigChangedEvent)
+
+
 class YourCharm(CharmBase):
+
+    on = KubernetesMultusCharmEvents()
 
     def __init__(self, *args):
         super().__init__(*args)
@@ -36,6 +50,7 @@ class YourCharm(CharmBase):
                 )
             ],
             network_attachment_definitions_func=self._network_attachment_definitions_from_config,
+            refresh_event=self.on.nad_config_changed,
         )
 
     def _network_attachment_definitions_from_config(self) -> list[NetworkAttachmentDefinition]:
@@ -66,6 +81,14 @@ class YourCharm(CharmBase):
                 },
             ),
         ]
+
+    def _on_config_changed(self, event: EventBase):
+        if not self.unit.is_leader():
+            return
+        if self._get_invalid_configs():
+            return
+        # Fire the NadConfigChangedEvent if the configs are valid.
+        self.on.nad_config_changed.emit()
 ```
 """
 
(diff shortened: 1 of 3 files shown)


lib/charms/kubernetes_charm_libraries/v0/multus.py
```diff
@@ -468,6 +502,7 @@ def __init__(
         container_name: str,
         cap_net_admin: bool = False,
         privileged: bool = False,
+        refresh_event: Optional[Union[BoundEvent, List[BoundEvent]]] = None,
```
gruyaume, line 505: Why make it optional? My recommendation is to have it simply be of type BoundEvent. No need to support list of events at the moment, afaik there is no use case for this right now. This should make the rest of the init quite simple:

```python
self.framework.observe(refresh_event, self._configure_multus)
```

lib/charms/kubernetes_charm_libraries/v0/multus.py
```diff
@@ -619,3 +662,7 @@ def _on_remove(self, event: RemoveEvent) -> None:
                 self.kubernetes.delete_network_attachment_definition(
                     name=network_attachment_definition.metadata.name  # type: ignore[union-attr]
                 )
+
+    def delete_pod(self) -> None:
+        """Delete the pod."""
+        self.kubernetes.delete_pod(self._pod)
```
gruyaume, line 668: Should this change be part of this PR?

lib/charms/kubernetes_charm_libraries/v0/multus.py
```diff
@@ -148,6 +160,15 @@ def __init__(self, namespace: str):
...
 
+    def delete_pod(self, pod_name: str) -> None:
+        """Deleting given pod.
+
+        Args:
+            pod_name (str): Pod name
+
+        """
+        self.client.delete(Pod, pod_name, namespace=self.namespace)
+
```
gruyaume, line 171: Should this change be part of this PR?

lib/charms/kubernetes_charm_libraries/v0/multus.py
```diff
@@ -457,6 +478,19 @@ def _container_security_context_is_set(
...
+    def __init__(self, handle: Handle):
+        super().__init__(handle)
+
+
+class NadConfigChangedCharmEvents(CharmEvents):
+    """NAD config changed events."""
+
+    nad_config_changed = EventSource(NadConfigChangedEvent)
+
+
```
gruyaume, line 493: I'm not sure we need to have this at all, users should be able to pass any event (ex. ConfigChanged or a custom event) and our code should not care.

gruyaume (changes requested):

lib/charms/kubernetes_charm_libraries/v0/multus.py
gatici: This belongs to another PR, it should be removed.

lib/charms/kubernetes_charm_libraries/v0/multus.py
gatici: This belongs to another PR, it should be removed.

lib/charms/kubernetes_charm_libraries/v0/multus.py
gatici: There is a reason to define the a custom events in the library part. The event to be observed will be a custom event as native charm events does not appropriate for our situation. If we do not define any custom event in the library side, each consumer charm needs to define custom events to be passed and observed. This requires more modifications and repeated work in the several consumer charms. 
If a custom event exist in the library part which is created for this purpose, charms can import and use it easily.

lib/charms/kubernetes_charm_libraries/v0/multus.py
gatici: I tried to support the single Bound event or list of Bound events by considering the future requirements. At the moment, single event is enough. We can support list of events when we need.

lib/charms/kubernetes_charm_libraries/v0/multus.py
gatici: This is removed from the library side.

gatici:
LXD version will be updated to 5.17/stable.

DanielArndt (changes requested):

DanielArndt:
Just the one small comment. Otherwise LGTM.

src/charm.py
```diff
@@ -60,9 +60,24 @@ class IncompatibleCPUError(Exception):
     pass
 
 
+class NadConfigChangedEvent(EventBase):
+    """Event triggered when an existing network attachment definition is changed."""
+
+    def __init__(self, handle: Handle):
+        super().__init__(handle)
```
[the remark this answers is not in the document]

gruyaume (approved):

DanielArndt (approved):
Ahh I guess with a docstring you don't even need `pass`. TIL.

gatici:
> Ahh I guess with a docstring you don't even need `pass`. TIL.

I will remove the init method of NadConfigChangedEvent from Docstring in the library part.

PR #52, opened 2023-10-27
fix: Prevent scaling by blocking on non-leader units

DanielArndt:
Prevent scaling by blocking on non-leader units.

Also, add note about best-effort removal of the k8s service.

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -95,7 +95,7 @@ class UPFOperatorCharm(CharmBase):
             # NOTE: In cases where leader status is lost before the charm is
             # finished processing all teardown events, this prevents teardown
             # event code from running. Luckily, for this charm, none of the
-            # teardown code is necessary to preform if we're removing the
+            # teardown code is necessary to perform if we're removing the
             # charm.
             self.unit.status = BlockedStatus("Scaling is not implemented for this charm")
             return
@@ -186,7 +186,7 @@ class UPFOperatorCharm(CharmBase):
         self._delete_external_upf_service()
 
     def _delete_external_upf_service(self) -> None:
-        # NOTE: We want to preform this removal only if the last remaining unit
+        # NOTE: We want to perform this removal only if the last remaining unit
         # is removed. This charm does not support scaling, so it *should* be
         # the only unit.
         #
--- a/src/charm.py
+++ b/src/charm.py
@@ -91,6 +91,14 @@ class UPFOperatorCharm(CharmBase):
 
     def __init__(self, *args):
         super().__init__(*args)
+        if not self.unit.is_leader():
+            # NOTE: In cases where leader status is lost before the charm is
+            # finished processing all teardown events, this prevents teardown
+            # event code from running. Luckily, for this charm, none of the
+            # teardown code is necessary to preform if we're removing the
+            # charm.
+            self.unit.status = BlockedStatus("Scaling is not implemented for this charm")
+            return
         self._bessd_container_name = self._bessd_service_name = "bessd"
         self._routectl_service_name = "routectl"
         self._pfcp_agent_container_name = self._pfcp_agent_service_name = "pfcp-agent"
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -91,6 +91,14 @@ class UPFOperatorCharm(CharmBase):
 
     def __init__(self, *args):
         super().__init__(*args)
+        if not self.unit.is_leader():
+            # NOTE: In cases where leader status is lost before the charm is
+            # finished processing all teardown events, this prevents teardown
+            # event code from running. Luckily, for this charm, none of the
+            # teardown code is necessary to preform if we're removing the
```
gatici, line 98:
```suggestion
            # teardown code is necessary to perform if we're removing the
```

src/charm.py
```diff
@@ -178,6 +186,21 @@ def _on_remove(self, event: RemoveEvent) -> None:
         self._delete_external_upf_service()
 
     def _delete_external_upf_service(self) -> None:
+        # NOTE: We want to preform this removal only if the last remaining unit
```
gatici, line 189:
```suggestion
        # NOTE: We want to perform this removal only if the last remaining unit
```

gatici (approved):
LGTM

src/charm.py
```diff
@@ -91,6 +91,14 @@ class UPFOperatorCharm(CharmBase):
 
     def __init__(self, *args):
         super().__init__(*args)
+        if not self.unit.is_leader():
+            # NOTE: In cases where leader status is lost before the charm is
+            # finished processing all teardown events, this prevents teardown
+            # event code from running. Luckily, for this charm, none of the
+            # teardown code is necessary to preform if we're removing the
+            # charm.
+            self.unit.status = BlockedStatus("Scaling is not implemented for this charm")
```
saltiyazan, line 100: I'm not sure about the status, blocked status indicates that an intervention from the user is required.
And let's assume having units 1 and 2, unit 1 is leader and active, unit 2 is non-leader and blocked, if leadership changes, will both of them go to the correct status? 1 being blocked and 2 being active? (I think yes, but asking to confirm :) )
And what if the condition mentioned in the comment changes and we need to run the teardown code when removing the charm?

saltiyazan: Anyway if this was the conclusion of the ongoing discussion I don't oppose it.

src/charm.py
DanielArndt: I have no idea why, but I always make this typo :laughing: thanks.

src/charm.py
DanielArndt: I share your concerns. :laughing: I don't like this approach at all, but there doesn't seem to be a great way to handle this.

> if leadership changes, will both of them go to the correct status? 1 being blocked and 2 being active?

Status will change correctly, although admittedly I'm unsure how well this well be handled without any of the set-up code being run. Blocked status is used, I guess, to indicate the the user should just un-scale their application. This was the best pattern @ghislainbourgeois and I could come up with while discussing this issue. Perhaps `Waiting` would be a better status? I'm unsure.

Additionally, we may be able to glean some patterns or a more robust way of handling this from one of these:

- [This discussion in charmhub](https://chat.charmhub.io/charmhub/pl/ze3qfn3yxirq7c4p4dyuo9k4bw)
- This code which looks at a related issue: https://github.com/canonical/mysql-router-k8s-operator/blob/876b2564d7397a1a8573fa989678766220664487/src/lifecycle.py#L128
- Holsim pattern: [Holism: let's make it official?](https://discourse.charmhub.io/t/holism-lets-make-it-official/12059/2) 
- This related issue: [Bug #1979811 “juju emitting `relation_broken` while there are re...” : Bugs : Canonical Juju](https://bugs.launchpad.net/juju/+bug/1979811) 
- The app.planned_units() function
- [This comment from Harry Pidcock](https://chat.charmhub.io/charmhub/pl/9847ybri4pdimqjdozurnk58bw)

I think we could come up with a pretty solid way to handle this situation, but it would be quite a bit of investigation and testing. That being said, I don't know how much time we want to spend digging on this, if we do eventually plan to support scaling.

saltiyazan: Put like that I see the point of the blocked status. Thanks.

src/charm.py
DanielArndt: Fixed in https://github.com/canonical/sdcore-upf-operator/pull/52/commits/7083a54806325bb35a6545c17705460ff3a0856b

src/charm.py
DanielArndt: Fixed in https://github.com/canonical/sdcore-upf-operator/pull/52/commits/7083a54806325bb35a6545c17705460ff3a0856b

DanielArndt:
@saltiyazan I'll leave this open since you mentioned in stand-up you may want to check over the concern regarding data bags.

Unfortunately, this same concern exists across the board on all our charms, so I'm almost certain there are edge cases in at least one of them. I don't think this is a long term fix, it's just "better than before".

DanielArndt:
FWIW the main goal of this task was to get the integration tests working more reliably, which I think they are. It would have been nice to have a high quality fix for this, but I didn't want to spiral for too long if it's not a priority.

saltiyazan (approved):
Based on your replies @DanielArndt I'm approving this.
But if we plan to implement this across all of the SD-Core charms I'd ask for one more eye from the team to take a look.

DanielArndt:
@saltiyazan It's already implemented across all the other charms, this was the last one :laughing: -- so, I guess that's to say others have looked at it as well.

Definitely appreciate your input concern, though, as I feel similar about it.

DanielArndt:
Definitely happy to revisit this if we want to invest more time, but this at least makes things slightly better...

PR #57, opened 2023-11-16
feat: Adding DPDK support in UPF

Gmerold:
# Description

This PR adds DPDK support to UPF.

Most important changes:
1. NADs creation has been refactored to support `vfioveth` type required by DPDK
2. StatefulSet network annotations creation has been refactored
3. NetworkAnnotation dataclass has 2 additional fields required by DPDK: `mac` and `ips`
4. `DPDKStatefulSetUpdater` has been added to handle StatefulSet updates required to run UPF in DPDK mode
5. New charm configs added

Design decissions:
1. `DPDKStatefulSetUpdater` is not a library, because at this point we are not aware of anyone else who could potentially leverage such a library. Regular class seems good enough at this point.
2. `DPDKStatefulSetUpdater` makes `bess` container a privileged container. That's because in DPDK mode, bess needs to access physical address of HugePages. Ideally `IPC_LOCK` capability should be enough, but it's not. [Similar issue](https://bugzilla.redhat.com/show_bug.cgi?id=1785933) has been described here.

# Checklist:


lib/charms/kubernetes_charm_libraries/v0/multus.py
ghislainbourgeois, line 1: Shouldn't all those changes be done in this project: https://github.com/canonical/kubernetes-charm-libraries?

src/charm.py
```diff
@@ -406,6 +532,12 @@ def _on_pfcp_agent_pebble_ready(self, event: EventBase) -> None:
         """Handle pfcp agent Pebble ready event."""
         if not self.unit.is_leader():
             return
+        if invalid_configs := self._get_invalid_configs():
+            self.unit.status = BlockedStatus(
+                f"The following configurations are not valid: {invalid_configs}"
+            )
+            event.defer()
```
ghislainbourgeois, line 539: I do not think the defer here is necessary, as `config_changed` will be emitted before this is fixed.

src/dpdk_statefulset_updater.py
```diff
@@ -0,0 +1,175 @@
...
+        self.dpdk_resource_requirements = {
+            "requests": {
+                dpdk_access_interface_resource_name: "1",
+                dpdk_core_interface_resource_name: "1",
+            },
+            "limits": {
+                dpdk_access_interface_resource_name: "1",
+                dpdk_core_interface_resource_name: "1",
+            },
+        }
```
ghislainbourgeois, line 48: I am not sure I understand those requirements, what do they represent exactly?

ghislainbourgeois (changes requested):

lib/charms/kubernetes_charm_libraries/v0/multus.py
Gmerold: Damn, I knew I'll eventually forget to move this :D

src/charm.py
Gmerold: Not exactly sure what you mean here. If we don't defer, then how do we make sure pfcp actually starts after the configs are fixed? We run this code only on `pfcp_agent_pebble_ready`. 
Should we run pfcp configuration on `config_changed` as well (as we already do for bess)?

src/dpdk_statefulset_updater.py
Gmerold: Network resources (VFs) are advertised to Kubernetes in the form of allocatable resources. We use [SR-IOV Device Plugin](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin) to map the resource to physical PCI address. For configured cluster it should look similar to this:
```
$ kubectl get node  -o json | jq '.items[].status.allocatable'
{
  "cpu": "7",
  "ephemeral-storage": "46779129369",
  "hugepages-1Gi": "2Gi",
  "hugepages-2Mi": "0",
  "intel.com/intel_sriov_vfio_access": "1",
  "intel.com/intel_sriov_vfio_core": "1",
  "memory": "13198484Ki",
  "pods": "110"
}
```
Container's ResourceRequirements are the way to reflect the fact that we need these devices to be available.

lib/charms/kubernetes_charm_libraries/v0/multus.py
Gmerold: Here's the [PR to update the lib](https://github.com/canonical/kubernetes-charm-libraries/pull/23).

requirements.txt
```diff
@@ -1,7 +1,11 @@
+dataclass-wizard
```
ghislainbourgeois, line 1: Following the changes in the lib, the requirements here will also need to change.

src/dpdk_statefulset_updater.py
ghislainbourgeois: Ok, so "1" is treated as a boolean in that case by the plugin. Thanks for clarifying!

src/charm.py
ghislainbourgeois: Looking a bit more at the code, I think we can just drop the checks for the invalid config in this hook, as the pfcp container does not use any configurations.

ghislainbourgeois (changes requested):

src/dpdk_statefulset_updater.py
Gmerold: Actually "1" represents quantity of resource of given type. In this case it means we have 1 interface of each type. Number of resources will depend on the number of PCI addresses assigned to a resource.

gruyaume (changes requested):

config.yaml
```diff
@@ -1,4 +1,10 @@
 options:
+  upf-mode:
+    type: string
+    default: af_packet
+    description: |
+      Either `af_packet` (default) or `dpdk`.
+      Please see README.md to learn about all required parameters.
```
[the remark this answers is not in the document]
Gmerold: Is there anything wrong with exposing to the user? The reason why I did it this way is that it corresponds nicely with the UPF config file, which has the `mode` key.
Not sure how to explain that, but using `boolean` for enabling DPDK mode seems kinda awkward to me :D It makes sense for HugePages - if you don't enable HugePages you just don't use HugePages. If you don't use DPDK, you use AF_PACKET. Does that make any sense?

lib/charms/kubernetes_charm_libraries/v0/multus.py
Gmerold: Done.

requirements.txt
Gmerold: Yes, removed.

src/charm.py
Gmerold: Indeed. Removed.

config.yaml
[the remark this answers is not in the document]
ghislainbourgeois: I also think it is better to expose it directly, as we might want to support other modes, like `af_xdp` in the future.

ghislainbourgeois (approved):

config.yaml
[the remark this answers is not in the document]
gruyaume: My concern is that "af_packet" doesn't mean much. How do I know which mode to choose? In any case I'm fine with keeping it as is and change it only if it's confusing users.

requirements.txt
```diff
@@ -1,7 +1,10 @@
+httpx
```
gruyaume, line 1: Where is this used?

src/dpdk_statefulset_updater.py
```diff
@@ -0,0 +1,175 @@
...
+
+import logging
+from typing import Iterable, Optional
+
+from lightkube import Client
+from lightkube.core.exceptions import ApiError
+from lightkube.models.core_v1 import Container
+from lightkube.resources.apps_v1 import StatefulSet
+
+logger = logging.getLogger(__name__)
```
gruyaume, line 15: We set this here but never use it. We should probably log after we completed some actions, example after the `replace` command.

src/dpdk_statefulset_updater.py
```diff
@@ -0,0 +1,175 @@
...
+
+class DPDKStatefulSetUpdaterError(Exception):
+    """DPDKStatefulSetUpdaterError."""
+
+    def __init__(self, message: str):
+        self.message = message
+        super().__init__(self.message)
+
+
+class DPDKStatefulSetUpdater:
```
gruyaume, line 26:
```suggestion
class DPDKStatefulSet:
```

src/dpdk_statefulset_updater.py
```diff
@@ -0,0 +1,175 @@
...
+                dpdk_access_interface_resource_name: "1",
+                dpdk_core_interface_resource_name: "1",
+            },
+            "limits": {
+                dpdk_access_interface_resource_name: "1",
+                dpdk_core_interface_resource_name: "1",
+            },
+        }
+
+    def container_configured_for_dpdk(self, container_name: str) -> bool:
```
gruyaume, line 50:
```suggestion
    def is_configured(self, container_name: str) -> bool:
```

src/dpdk_statefulset_updater.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+        if not container:
+            raise RuntimeError("Container not found!")
+        if not container.securityContext.privileged:
+            return False
+        if not self._resource_requirements_applied(container, self.dpdk_resource_requirements):
+            return False
+        return True
+
+    def configure_container_for_dpdk(self, container_name: str) -> None:
```
gruyaume, line 74:
```suggestion
    def configure(self, container_name: str) -> None:
```

src/dpdk_statefulset_updater.py
gruyaume: There's already dpdk in the class name, it's self explanatory.

src/dpdk_statefulset_updater.py
```diff
@@ -0,0 +1,175 @@
...
+        """Applies config required by DPDK to a given container.
+
+        Args:
+            container_name (str): Name of the container to update
+        """
+        statefulset = self._get_statefulset(self.statefulset_name, self.namespace)
+        if not statefulset:
+            raise RuntimeError("StatefulSet not found!")
```
gruyaume, line 82: Why do we raise RuntimeError instead of DPDKStatefulSetUpdaterError

tests/unit/test_charm.py
```diff
@@ -612,17 +700,333 @@ def test_given_default_config_with_interfaces_when_network_attachment_definition
...
-            self.assertEqual(config["master"], nad.metadata.name)
+            self.assertTrue(ACCESS_INTERFACE_NAME or CORE_INTERFACE_NAME in config["master"])
             self.assertEqual(config["type"], "macvlan")
 
+    def test_given_upf_configured_to_run_in_dpdk_mode_when_network_attachment_definitions_from_config_is_called_then_2_nads_are_returned(  # noqa: E501
+        self,
+    ):
+        self.harness.disable_hooks()
+        self.harness.update_config(key_values={"upf-mode": "dpdk"})
+        nads = self.harness.charm._network_attachment_definitions_from_config()
```
gruyaume, line 719: I know we do this at other places in the unit tests but I don't think it's a good idea to test this private method specifically. If we think it's worth testing on its own, I'd argue it should be in a class of its own (or be a method outside of this class) but since it's not the case, we are validating our implementation of the code and making our tests more brittle.

tests/unit/test_charm.py
```diff
@@ -612,17 +700,333 @@ def test_given_default_config_with_interfaces_when_network_attachment_definition
...
+        self.harness.disable_hooks()
+        self.harness.update_config(key_values={"upf-mode": "dpdk"})
+        network_annotations = self.harness.charm._generate_network_annotations()
+        self.assertEqual(len(network_annotations), 2)
+
+    def test_given_upf_charm_configured_to_run_in_default_mode_when_generate_network_annotations_is_called_then_access_nad_annotation_created(  # noqa: E501
+        self,
+    ):
+        self.harness.disable_hooks()
+        network_annotations = self.harness.charm._generate_network_annotations()
```
gruyaume, line 766: Same comment here, why test those private calls?

gruyaume (changes requested):

requirements.txt
Gmerold: In the Multus lib.

src/dpdk_statefulset_updater.py
Gmerold: Because this should never happen. `_get_statefulset` will raise `DPDKStatefulSetUpdaterError` when appropriate. This check is in fact just to make static analysis happy.
Alternatively, I could do 
```python
assert statefulset
```

src/dpdk_statefulset_updater.py
Gmerold: How about something like `DPDKConfigurator`? It would look good with public methods called `is_configured` and `configure` you suggested below.

src/dpdk_statefulset_updater.py
Gmerold: True, added.

src/dpdk_statefulset_updater.py
Gmerold: Done.

src/dpdk_statefulset_updater.py
Gmerold: Done.

tests/unit/test_charm.py
Gmerold: True, we do it other places, but I'd say it's rather exceptional. In this case (and probably other cases as well), I did it like this to avoid a boatload of mocking and patching. Also in this case, I don't necessarily agree that it makes our tests brittle. These do not care do we build up our network annotations - in whatever way it is done, the tests only check whether the output contains what it should. 
I think the choice we have here is:
1. Mock and patch a lot of stuff, build up the harness, fire events, get the statefulset and validate network annotations, or
2. Call the function that builds network annotation and check the output.

And I actually think that moving the logic for building up the Network annotation outside the charm class makes a lot of sense.

gruyaume: It makes our tests brittle in the sense that they can break even though the behavior is maintainted. Specifically, if w remove `_network_attachment_definitions_from_config` but keep the same behavior implemented differently, we find ourselves with 20 broken tests.

src/dpdk_statefulset_updater.py
Gmerold: Cool, I like that.
Done.

tests/unit/test_charm.py
Gmerold: I agree, but on the other hand if we need to patch most of the landscape just to test the creation of Network annotations, we may end up with 20 broken tests as well if we, for instance, change the lib we use to talk to Kubernetes. 
Personally, if I need to, I'd rather fix 20 trivial tests than 20 tests full of patches.
I'm not saying we should start testing all private methods, but I think there are situations where it is justified.

gruyaume: Those libs are public classes and methods, yes we expect to change the tests if those change. Here `_network_attachment_definitions_from_config` is private and purely an implementation detail.

Gmerold: You are one tough adversary in this discussion :rofl: :rofl: 
I give up. I wrote new tests, which do not call private methods.

tests/unit/test_charm.py
```diff
@@ -612,17 +701,546 @@ def test_given_default_config_with_interfaces_when_network_attachment_definition
...
+    @patch("lightkube.core.client.Client.create")
+    @patch("ops.model.Container.get_service")
+    @patch("lightkube.core.client.GenericSyncClient", new=Mock)
+    @patch(
+        f"{HUGEPAGES_LIBRARY_PATH}.KubernetesHugePagesPatchCharmLib.is_patched",
+        Mock(return_value=True),
+    )
+    @patch(f"{MULTUS_LIBRARY_PATH}.KubernetesMultusCharmLib.is_ready", Mock(return_value=True))
+    @patch("charm.DPDK.is_configured", Mock(return_value=True))
+    @patch("ops.model.Container.exec", new=MagicMock)
```
gruyaume, line 785: We likely don't need to mock `exec` calls anymore as of PR #59

gruyaume (approved):

gatici:
After deploying the charm with DPDK mode, bessd container can not start with following error.
```
2023-11-28T19:58:45.592Z [pebble] Service "bessd" starting: /bin/bessd -f -grpc-url=0.0.0.0:10514 
2023-11-28T19:58:45.706Z [bessd] F1128 19:58:45.706557    53 bessd.cc:211] flock(pidfile=/var/run/bessd.pid): Permission denied [13]
2023-11-28T19:58:45.706Z [bessd] *** Check failure stack trace: ***
2023-11-28T19:58:45.727Z [bessd] F1128 19:58:45.727789    53 debug.cc:407] Backtrace (recent calls first) ---
2023-11-28T19:58:45.727Z [bessd] (0): /bin/bessd(_ZN6google15ErrnoLogMessageD1Ev+0xae) [0x55de3962323e]
2023-11-28T19:58:45.727Z [bessd] (1): /bin/bessd(_ZN4bess5bessd21TryAcquirePidfileLockEi+0x163) [0x55de38cf6863]
2023-11-28T19:58:45.727Z [bessd] (2): /bin/bessd(_ZN4bess5bessd19CheckUniqueInstanceERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x49) [0x55de38cf6989]
2023-11-28T19:58:45.727Z [bessd] (3): /bin/bessd(main+0xe3) [0x55de38a65d73]
2023-11-28T19:58:45.727Z [bessd] (4): /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf1) [0x7f92cd8f5081]
2023-11-28T19:58:45.727Z [bessd] (5): /bin/bessd(_start+0x2c) [0x55de38cbbefc]
2023-11-28T19:58:45.727Z [bessd] *** Check failure stack trace: ***
```

src/charm.py
```diff
@@ -124,16 +129,7 @@ def __init__(self, *args):
             charm=self,
             container_name=self._bessd_container_name,
             cap_net_admin=True,
```
gatici, line 131: Does bessd container require more privilege other than NET_ADMIN ?

Gmerold:
> After deploying the charm with DPDK mode, bessd container can not start with following Permission Deny error.
> 
> ```
> 2023-11-28T19:58:45.592Z [pebble] Service "bessd" starting: /bin/bessd -f -grpc-url=0.0.0.0:10514 
> 2023-11-28T19:58:45.706Z [bessd] F1128 19:58:45.706557    53 bessd.cc:211] flock(pidfile=/var/run/bessd.pid): Permission denied [13]
> 2023-11-28T19:58:45.706Z [bessd] *** Check failure stack trace: ***
> 2023-11-28T19:58:45.727Z [bessd] F1128 19:58:45.727789    53 debug.cc:407] Backtrace (recent calls first) ---
> 2023-11-28T19:58:45.727Z [bessd] (0): /bin/bessd(_ZN6google15ErrnoLogMessageD1Ev+0xae) [0x55de3962323e]
> 2023-11-28T19:58:45.727Z [bessd] (1): /bin/bessd(_ZN4bess5bessd21TryAcquirePidfileLockEi+0x163) [0x55de38cf6863]
> 2023-11-28T19:58:45.727Z [bessd] (2): /bin/bessd(_ZN4bess5bessd19CheckUniqueInstanceERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x49) [0x55de38cf6989]
> 2023-11-28T19:58:45.727Z [bessd] (3): /bin/bessd(main+0xe3) [0x55de38a65d73]
> 2023-11-28T19:58:45.727Z [bessd] (4): /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf1) [0x7f92cd8f5081]
> 2023-11-28T19:58:45.727Z [bessd] (5): /bin/bessd(_start+0x2c) [0x55de38cbbefc]
> 2023-11-28T19:58:45.727Z [bessd] *** Check failure stack trace: ***
> ```

Is this strictly confined Microk8s that you're running there? If yes, it won't work.

tests/unit/test_charm.py
Gmerold: True. Rebased and cleaned up.

gatici (approved):

README.md
```diff
@@ -32,7 +32,23 @@ juju deploy sdcore-upf --trust --channel=edge
...
+If a load balancer such as `metallb` is present, the charm will configure an externally accessible 
+service port with the load balancer upon install of the charm.
+
+### Running UPF in DPDK mode
+
+By default, UPF runs in `af_packet` mode. To run UPF in `dpdk` mode, `upf-mode` config option 
+should be used, i.e.:
+
+```shell
+juju deploy sdcore-upf --trust --channel=edge --config upf-mode="dpdk" --config enable-hugepages=True --config access-interface-mac-address="00:b0:d0:63:c2:26" --config core-interface-mac-address="00:b0:d0:63:c2:36"
```
gatici, line 44: As we discussed earlier, we need to add a  note that  users need to set following configs according to their environment:
 - core-ip
 - core-gateway-ip
 - access-ip
 - access-gateway-ip

Gmerold: I've added the note, but as always, in this kind of cases, I'm not sure about the phrasing :)
@gruyaume would you mind taking a look? Thanks!

config.yaml
gruyaume, line 7: The descriptions here are accessible to users when they run `juju config`, not the readme. Instead of pointing to the README, it's preferable to place all the information needed to configure dpdk here. Ex. the fact that we need hugepages.

README.md
gruyaume: I think the note is fine, my concern really is more about where this should be documented. I highly prefer placing any requirement in `config.yaml`, closer to the user. See my other comment below.

config.yaml
Gmerold: I removed this reference. Now all configs required by the DPDK are described accordingly.

PR #59, opened 2023-11-17
chore: Use handle_exec instead of mocking container exec method

dariofaccin:
# Description

This PR aims to replace mock objects for `ops.model.Container.exec`, leveraging the new `ops.harness.handle_exec` method.

# Checklist:

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -3,7 +3,8 @@
 
 import json
 import unittest
-from unittest.mock import MagicMock, Mock, call, patch
+from dataclasses import dataclass
+from unittest.mock import Mock, call, patch
 
 from charms.kubernetes_charm_libraries.v0.multus import (  # type: ignore[import]
     NetworkAttachmentDefinition,
@@ -13,7 +14,6 @@ from lightkube.models.meta_v1 import ObjectMeta
 from lightkube.resources.core_v1 import Service
 from ops import testing
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
-from ops.pebble import ExecError
 
 from charm import IncompatibleCPUError, UPFOperatorCharm
 
@@ -33,6 +33,13 @@ ACCESS_GW_IP = "192.168.252.1"
 GNB_SUBNET = "192.168.251.0/24"
 CORE_IP = "192.168.250.3/24"
 CORE_GW_IP = "192.168.250.1"
+
+
+@dataclass
+class HandleExecCallArgs:
+    command: list
+    timeout: float | None
+    environment: dict
 
 
 def read_file(path: str) -> str:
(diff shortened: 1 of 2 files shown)


tests/unit/test_charm.py
```diff
@@ -168,71 +175,120 @@ def test_given_when_bessd_pebble_ready_then_expected_pebble_plan_is_applied(  #
...
+            call_list.append(
+                HandleExecCallArgs(
+                    command=args.command,
+                    timeout=args.timeout,
+                    environment=args.environment,
+                )
+            )
+            return testing.ExecResult(exit_code=0)
+
+        self.harness.handle_exec("bessd", [], handler=bessd_cmds_handler)
```
gruyaume, line 193: @benhoyt is this the recommended approach to use `handle_exec`?

benhoyt: Not exactly. This looks like a direct port of the existing patch code, which is fine. But I'd recommend using the command matching part of `handle_exec`. For example:

```python
    @patch(f"{MULTUS_LIBRARY_PATH}.KubernetesMultusCharmLib.is_ready")
    def test_given_can_connect_to_bessd_when_bessd_pebble_ready_then_ip_route_is_created(
        self, patch_is_ready
    ):
        ip_command = ["ip", "route", "replace", "default", "via", CORE_GW_IP, "metric", "110"]
        self.harness.handle_exec("bessd", ip_command, result=0)
        self.harness.handle_exec("bessd", ["iptables-legacy"], result=0)
        self.harness.handle_exec("bessd", ["/opt/bess/bessctl/bessctl"], result=0)

        patch_is_ready.return_value = True

        self.harness.container_pebble_ready(container_name="bessd")
```

Or if you really want to ensure that the `ip route ...` command was called at least once:

```python
    @patch(f"{MULTUS_LIBRARY_PATH}.KubernetesMultusCharmLib.is_ready")
    def test_given_can_connect_to_bessd_when_bessd_pebble_ready_then_ip_route_is_created(
        self, patch_is_ready
    ):
        ip_route_called = False

        def ip_handler(args: testing.ExecArgs) -> testing.ExecResult:
            nonlocal ip_route_called
            ip_route_called = True
            return testing.ExecResult(exit_code=0)

        ip_command = ["ip", "route", "replace", "default", "via", CORE_GW_IP, "metric", "110"]
        self.harness.handle_exec("bessd", ip_command, handler=ip_handler)
        self.harness.handle_exec("bessd", ["iptables-legacy"], result=0)
        self.harness.handle_exec("bessd", ["/opt/bess/bessctl/bessctl"], result=0)

        patch_is_ready.return_value = True

        self.harness.container_pebble_ready(container_name="bessd")

        self.assertTrue(ip_route_called)
```

That second example is not all that far from what you have in the PR now (you're testing for a side effect), but it's better to use a more specific command prefix in the `handle_exec` call.

dariofaccin: Thanks for the precious hints. I updated the PR, now it should be better.

gruyaume (approved):

PR #67, opened 2023-12-07
feat: Set IP masquerade in the UPF configuration

ghislainbourgeois:
# Description

Sets IP masquerade on the core interface of the UPF. This will automatically NAT UE traffic to appear to come from the UPF's core interface, making it simpler to integrate in existing networks.

The alternative is to have the core router know about the UE IP pool to be able to route back traffic towards the UPF.

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -411,12 +411,13 @@ class UPFOperatorCharm(CharmBase):
         Writes configuration file, creates routes, creates iptable rule and pebble layer.
         """
         restart = False
+        core_ip_address = self._get_core_network_ip_config()
         content = render_bessd_config_file(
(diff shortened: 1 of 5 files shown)


markbeierl:
Are we doing this unconditionally or should we expose an option not to do it?

ghislainbourgeois:
> Are we doing this unconditionally or should we expose an option not to do it?

As it is, it is unconditional. I think for now this can be enough, but we might have to add an option in the future most likely. I could be convinced of adding the option right away, but would probably put the default value to masquerade.

WDYT?

markbeierl (approved):

PR #69, opened 2023-12-11
fix: move to Blocked state when HugePages are not available

dariofaccin:
# Description

This PR adds two validations to the UPF charm when enabling HugePages: the first is the support of the required CPU flags (`pdpe1gb`); the second is the capacity of enough HugePages (2Gi) on the K8S nodes. If the free HugePages (allocatable minus allocated) are not enough, the charm will remain in Waiting status.

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -55,6 +55,7 @@ BESSD_PORT = 10514
 PROMETHEUS_PORT = 8080
 PFCP_PORT = 8805
 REQUIRED_CPU_EXTENSIONS = ["avx2", "rdrand"]
+REQUIRED_CPU_EXTENSIONS_HUGEPAGES = ["pdpe1gb"]
 
 # The default field manager set when using kubectl to create resources
 DEFAULT_FIELD_MANAGER = "controller"
@@ -367,6 +368,13 @@ class UPFOperatorCharm(CharmBase):
         """Handler for config changed events."""
         if not self.unit.is_leader():
             return
+        if self._hugepages_is_enabled():
+            if not self._cpu_is_compatible_for_hugepages():
+                raise IncompatibleCPUError(
+                    "\nCPU is not compatible!\n"
+                    "Please use a CPU that has the following capabilities: "
+                    f"{', '.join(REQUIRED_CPU_EXTENSIONS + REQUIRED_CPU_EXTENSIONS_HUGEPAGES)}"
+                )
         if invalid_configs := self._get_invalid_configs():
             self.unit.status = BlockedStatus(
                 f"The following configurations are not valid: {invalid_configs}"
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -367,6 +368,16 @@ def _on_config_changed(self, event: EventBase):
...
             return
+        if self._hugepages_is_enabled():
+            if not self._cpu_is_compatible_for_hugepages():
+                raise IncompatibleCPUError(
+                    "\nCPU is not compatible!\n"
+                    "Please use a CPU that has the following capabilities: "
+                    f"{', '.join(REQUIRED_CPU_EXTENSIONS + REQUIRED_CPU_EXTENSIONS_HUGEPAGES)}"
+                )
+            if not self._hugepages_are_available():
+                self.unit.status = BlockedStatus("Not enough HugePages available")
```
gruyaume, line 379: I think we should also observe `update_status` because no event may be triggered when finally hugepages are available

dariofaccin: This method checks that at least 2Gi of HugePages are allocatable. For this value to change, the user should either restart the host or the k8s node, causing the pod to also restart.

src/charm.py
```diff
@@ -788,6 +799,20 @@ def _get_cpu_extensions() -> list[str]:
...
+            for required_extension in REQUIRED_CPU_EXTENSIONS_HUGEPAGES
+        )
+
+    @staticmethod
+    def _hugepages_are_available() -> bool:
+        client = Client()
+        nodes = client.list(Node)
+        if not nodes:
+            return False
+        return all([node.status.capacity.get("hugepages-1Gi", "0") >= "2Gi" for node in nodes])
```
Gmerold, line 814: I'd suggest using `allocatable` instead of `capacity`. Some part of `capacity` will be reserved by the system, kubelet, etc.

Gmerold (changes requested):
PR description says:
`(...) If the free HugePages (allocatable minus allocated) are not enough, the charm will remain in Waiting status.`
The code only checks the capacity tho. I think it would be great to have the feature implemented as described above.

src/charm.py
gatici: I think checking node capacity is fine here. Because If the node does not have 2Gi capacity, the charm will go to Blocked state.  However, the mode may have capacity but not enough resource. In this case, the bessd container will not start and the pod will be in waiting status.

gatici (approved):

src/charm.py
Gmerold: Yes, in this case it's not critical, but in general, I'd rather use `allocatable`.

Gmerold (approved):

src/charm.py
dariofaccin: @gruyaume I can add `event.defer()` here, as in the worst case the event will be re-run when `update-status` is fired

gruyaume (approved):

PR #79, opened 2024-01-17
fix: Changing default value of hwcksum to true

Gmerold:
# Description

Changes the default value of `enable-hw-checksum` to True. Setting it to False breaks the simulation.

# Checklist:

### DIFF
--- a/tests/unit/expected_upf.json
+++ b/tests/unit/expected_upf.json
@@ -14,7 +14,7 @@
   },
   "enable_notify_bess": true,
   "gtppsc": true,
-  "hwcksum": false,
+  "hwcksum": true,
   "log_level": "trace",
   "max_sessions": 50000,
   "measure_flow": false,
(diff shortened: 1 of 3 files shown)


tests/unit/expected_upf.json
```diff
@@ -14,7 +14,7 @@
   },
   "enable_notify_bess": true,
   "gtppsc": true,
-  "hwcksum": false,
+  "hwcksum": true,
```
ghislainbourgeois, line 17: When adding this, I added a test that when the config was changed to `True`, it updated the rendered config. It should probably be reversed now also.

ghislainbourgeois (changes requested):

tests/unit/expected_upf.json
Gmerold: Good catch. 
Done.

ghislainbourgeois (approved):

PR #83, opened 2024-02-02
feat: use Pydantic validator and support for new CNIs

dariofaccin:
# Description

This PR aims to leverage Pydantic to validate charm config input parameters.
Support for `host-device` CNI type is added.
Config option `enable-hugepages` is removed and it is inferred from `upf-mode`.

# Checklist:

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -2,6 +2,9 @@ options:
   cni-type:
     type: string
     default: bridge
+    description: |
+      Multus CNI plugin to use for the interfaces.
+      Allowed values are `bridge`, `host-device`, `macvlan`, `vfioveth`.
   upf-mode:
     type: string
     default: af_packet
--- a/config.yaml
+++ b/config.yaml
@@ -1,4 +1,7 @@
 options:
+  cni-type:
+    type: string
+    default: bridge
   upf-mode:
     type: string
     default: af_packet
@@ -61,12 +64,6 @@ options:
       If not provided, it will default to the LoadBalancer Service hostname. 
       If that is not available, it will default to the internal
       Kubernetes FQDN of the service.
-  enable-hugepages:
-    type: boolean
-    default: false
-    description: |
-      When enabled, HugePages of 1Gi will be used for a total of 2Gi HugePages.
-      HugePages must be enabled if `upf-mode` is `dpdk`.
   enable-hw-checksum:
     type: boolean
     default: true
--- a/src/charm.py
+++ b/src/charm.py
@@ -237,7 +237,7 @@ class UPFOperatorCharm(CharmBase):
 
     def _update_fiveg_n3_relation_data(self) -> None:
         """Publishes UPF IP address in the `fiveg_n3` relation data bag."""
-        upf_access_ip_address = self._get_network_ip_config("access").split("/")[0]  # type: ignore[union-attr]  # noqa: E501
+        upf_access_ip_address = self._get_network_ip_config(ACCESS_INTERFACE_NAME).split("/")[0]  # type: ignore[union-attr]  # noqa: E501
         fiveg_n3_relations = self.model.relations.get("fiveg_n3")
         if not fiveg_n3_relations:
             logger.info("No `fiveg_n3` relations found.")
@@ -303,10 +303,10 @@ class UPFOperatorCharm(CharmBase):
             interface=CORE_INTERFACE_NAME,
         )
         if self._get_upf_mode() == "dpdk":
-            access_network_annotation.mac = self._get_interface_mac_address("access")
-            access_network_annotation.ips = [self._get_network_ip_config("access")]
-            core_network_annotation.mac = self._get_interface_mac_address("core")
-            core_network_annotation.ips = [self._get_network_ip_config("core")]
+            access_network_annotation.mac = self._get_interface_mac_address(ACCESS_INTERFACE_NAME)
+            access_network_annotation.ips = [self._get_network_ip_config(ACCESS_INTERFACE_NAME)]
+            core_network_annotation.mac = self._get_interface_mac_address(CORE_INTERFACE_NAME)
+            core_network_annotation.ips = [self._get_network_ip_config(CORE_INTERFACE_NAME)]
         return [access_network_annotation, core_network_annotation]
 
     def _network_attachment_definitions_from_config(self) -> list[NetworkAttachmentDefinition]:
@@ -319,8 +319,8 @@ class UPFOperatorCharm(CharmBase):
             access_nad = self._create_dpdk_access_nad_from_config()
             core_nad = self._create_dpdk_core_nad_from_config()
         else:
-            access_nad = self._create_nad_from_config("access")
-            core_nad = self._create_nad_from_config("core")
+            access_nad = self._create_nad_from_config(ACCESS_INTERFACE_NAME)
+            core_nad = self._create_nad_from_config(CORE_INTERFACE_NAME)
 
         return [access_nad, core_nad]
 
@@ -356,7 +356,7 @@ class UPFOperatorCharm(CharmBase):
                 {
                     "bridge": (
                         ACCESS_INTERFACE_BRIDGE_NAME
-                        if interface_name == "access"
+                        if interface_name == ACCESS_INTERFACE_NAME
                         else CORE_INTERFACE_BRIDGE_NAME
                     )
                 }
@@ -367,7 +367,7 @@ class UPFOperatorCharm(CharmBase):
             metadata=ObjectMeta(
                 name=(
                     CORE_NETWORK_ATTACHMENT_DEFINITION_NAME
-                    if interface_name == "core"
+                    if interface_name == CORE_INTERFACE_NAME
                     else ACCESS_NETWORK_ATTACHMENT_DEFINITION_NAME
                 )
             ),
@@ -516,7 +516,7 @@ class UPFOperatorCharm(CharmBase):
         Writes configuration file, creates routes, creates iptable rule and pebble layer.
         """
         restart = False
-        core_ip_address = self._get_network_ip_config("core")
+        core_ip_address = self._get_network_ip_config(CORE_INTERFACE_NAME)
         content = render_bessd_config_file(
             upf_hostname=self._upf_hostname,
             upf_mode=self._get_upf_mode(),  # type: ignore[arg-type]
@@ -769,9 +769,9 @@ class UPFOperatorCharm(CharmBase):
         Returns:
             Optional[str]: The network IP address to use
         """
-        if interface_name == "access":
+        if interface_name == ACCESS_INTERFACE_NAME:
             return str(self._charm_config.upf_config.access_ip)
-        elif interface_name == "core":
+        elif interface_name == CORE_INTERFACE_NAME:
             return str(self._charm_config.upf_config.core_ip)
         else:
             return None
@@ -785,9 +785,9 @@ class UPFOperatorCharm(CharmBase):
         Returns:
             Optional[str]: The interface on the host to use
         """
-        if interface_name == "access":
+        if interface_name == ACCESS_INTERFACE_NAME:
             return self._charm_config.upf_config.access_interface
-        elif interface_name == "core":
+        elif interface_name == CORE_INTERFACE_NAME:
             return self._charm_config.upf_config.core_interface
         else:
             return None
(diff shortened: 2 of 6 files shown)


src/charm.py
```diff
@@ -38,6 +36,7 @@
 from ops.model import ActiveStatus, BlockedStatus, Container, ModelError, WaitingStatus
 from ops.pebble import ExecError, Layer
 
+from charm_state import CharmConfigInvalidError, CharmState
```
Gmerold, line 39: I thought we'll be initializing `CharmState` (or `CharmConfig` as mentioned in another comment) using `from_charm` in the constructor, to get the validated config and then we'll get values from there instead of `return self.model.config.get("cni-type")`. 
Is there any reason why we're not doing this?

Gmerold (changes requested):

src/charm_state.py
```diff
@@ -0,0 +1,159 @@
...
+
+class LaxIPvAnyNetwork(IPvAnyNetwork):
+    """Validate an IPv4 or IPv6 network."""
+
+    def __new__(cls, value: NetworkType):
+        """Validate an IPv4 or IPv6 network."""
+        # Assume IP Network is defined with a default value for `strict` argument.
+        # Define your own class if you want to specify network address check strictness.
+        try:
+            return IPv4Network(value, strict=False)
```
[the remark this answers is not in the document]
dariofaccin: This will always fail: if `value` is a valid IPv4 network, then calling IPv6Network will raise an exception.

src/charm.py
dariofaccin: It seems that using `self._charm_config = CharmConfig.from_charm(charm=self)` the the charm init won't work: at that point the `charm.config` reflects the values before updating the config. One solution could be to instantiate a `CharmConfig` instance on `config-changed` and `pebble-ready`. Wdyt?

src/charm_state.py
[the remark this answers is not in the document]
Gmerold: Ahh, right, it raises an error rather than returning False. Sorry, my bad.

src/charm.py
Gmerold: I'm not sure what you mean by `reflects the values before updating the config`? Every time an event happens, new instance of the charm is created. It seems to me, that when the constructor is called, it'll always read the current config of the charm, because it's the config change that triggers the event not the other way around.
Please correct me if I'm missing something.

src/charm.py
```diff
@@ -325,49 +329,48 @@ def _network_attachment_definitions_from_config(self) -> list[NetworkAttachmentD
...
-        else:
-            access_nad_config.update({"type": "bridge", "bridge": ACCESS_INTERFACE_BRIDGE_NAME})
-
-        return NetworkAttachmentDefinition(
-            metadata=ObjectMeta(name=ACCESS_NETWORK_ATTACHMENT_DEFINITION_NAME),
-            spec={"config": json.dumps(access_nad_config)},
-        )
-
-    def _create_core_nad_from_config(self) -> NetworkAttachmentDefinition:
+    def _create_nad_from_config(self, interface_name: str) -> NetworkAttachmentDefinition:
```
gatici, line 337: We need to add new argument to the method docstring part.

src/charm.py
```diff
@@ -408,39 +411,21 @@ def _create_dpdk_core_nad_from_config(self) -> NetworkAttachmentDefinition:
...
             "cniVersion": "0.3.1",
             "ipam": {
                 "type": "static",
             },
             "capabilities": {"mac": True},
         }
-        if core_mtu := self._get_core_interface_mtu_config():
-            base_core_nad_config.update({"mtu": core_mtu})
-        return base_core_nad_config
+        return base_nad_config
```
gatici, line 428: return {
            "cniVersion": "0.3.1",
            "ipam": {
                "type": "static",
            },
            "capabilities": {"mac": True},
        }

src/charm.py
```diff
@@ -847,143 +750,61 @@ def _routectl_environment_variables(self) -> dict:
...
+    def _get_interface_config(self, interface_name: str) -> Optional[str]:
+        if interface_name == "access":
+            return self._charm_config.upf_config.access_interface
+        elif interface_name == "core":
+            return self._charm_config.upf_config.core_interface
+        else:
+            return None
 
-    def _get_access_interface_mac_address(self) -> Optional[str]:
+    def _get_interface_mac_address(self, interface_name: str) -> Optional[str]:
```
gatici, line 782: Please add argument to the docstrings.

src/charm.py
```diff
@@ -847,143 +750,61 @@ def _routectl_environment_variables(self) -> dict:
...
+            return None
 
-        Returns:
-            bool: Whether the gnb-subnet config is valid
-        """
-        gnb_subnet = self._get_gnb_subnet_config()
-        if not gnb_subnet:
-            return False
-        return ip_is_valid(gnb_subnet)
+    def _get_network_gateway_ip_config(self, interface_name: str) -> Optional[str]:
```
gatici, line 795: Please add Docstrings to the method.

src/charm_config.py
```diff
@@ -0,0 +1,155 @@
...
+    """Class to define available UPF modes for UPF operator."""
+
+    af_packet = "af_packet"
+    dpdk = "dpdk"
+
+
+NetworkType: TypeAlias = "str | bytes | int | tuple[str | bytes | int, str | int]"
+
+
+class LaxIPvAnyNetwork(IPvAnyNetwork):
```
gatici, line 48: The class name did not make sense at the first glance.

config.yaml
```diff
@@ -1,4 +1,7 @@
 options:
+  cni-type:
+    type: string
+    default: bridge
```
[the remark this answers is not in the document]
gatici: Agree. We need add a description which summarizes supported CNI types.

gatici (changes requested):
Some minor  comments are added.  Unit tests needs to be fixed.

src/charm_config.py
dariofaccin: I used `Lax` as opposite of `Strict`. I can think of `Loose`. Do you have any suggestion?

patriciareinoso (commented):
`access` and `core` could be consts

src/charm_config.py
patriciareinoso: NonStrict ? I was also confused about the name

src/charm_config.py
```diff
@@ -0,0 +1,155 @@
...
+    core_interface_mtu_size: Optional[int] = Field(default=None, ge=1200, le=65535)
+    external_upf_hostname: Optional[StrictStr] = Field(default="")
+    vlan_id: Optional[int] = Field(default=None, ge=1, le=4095)
+    enable_hw_checksum: bool = True
+
+    @model_validator(mode="after")
+    @classmethod
+    def validate_upf_mode_with_mac_addresses(cls, values):
+        """Validate that MAC addresses are defined when in DPDK mode."""
+        if values.upf_mode == "dpdk":
```
patriciareinoso, line 114: should this be something like UpfMode.dpdk ?

src/charm_config.py
```diff
@@ -0,0 +1,155 @@
...
+    core_ip: LaxIPvAnyNetwork = LaxIPvAnyNetwork("192.168.250.3/24")
+    core_gateway_ip: IPvAnyAddress = IPvAnyAddress("192.168.250.1")
+    core_interface_mtu_size: Optional[int] = Field(default=None, ge=1200, le=65535)
+    external_upf_hostname: Optional[StrictStr] = Field(default="")
+    vlan_id: Optional[int] = Field(default=None, ge=1, le=4095)
+    enable_hw_checksum: bool = True
+
+    @model_validator(mode="after")
+    @classmethod
+    def validate_upf_mode_with_mac_addresses(cls, values):
```
patriciareinoso, line 112: according to https://docs.pydantic.dev/latest/concepts/validators/

for `mode='after'` validators You should not use (cls, ModelType) as the signature, instead just use (self)

patriciareinoso (changes requested):

patriciareinoso:
what do you think about taking advantage of the types `CNIType` and `UpfMode` inside charm.py ?

config.yaml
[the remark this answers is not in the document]
gatici: The description is still missing.

src/charm.py
```diff
@@ -311,53 +315,63 @@ def _network_attachment_definitions_from_config(self) -> list[NetworkAttachmentD
...
-        core_nad_config["ipam"].update(
-            {"addresses": [{"address": self._get_core_network_ip_config()}]}
+        nad_config = self._get_nad_base_config()
+        cni_type = self._get_cni_type_config()
+        # MTU is optional for bridge, macvlan, dpdk
+        # MTU is ignored by host-device
+        if cni_type == CNIType.host_device:
+            pass
+        else:
+            if interface_mtu := self._get_interface_mtu_config(interface_name):
```
gatici, line 343:
```suggestion
        if cni_type != CNIType.host_device:
            if interface_mtu := self._get_interface_mtu_config(interface_name):
```

src/charm.py
```diff
@@ -620,84 +614,17 @@ def _create_bessctl_executed_validation_file(self, content) -> None:
...
-                invalid_configs.append("access-interface-mac-address")
-            if not self._core_interface_mac_address_is_valid():
-                invalid_configs.append("core-interface-mac-address")
-        return invalid_configs
-
     def _create_default_route(self) -> None:
         """Creates ip route towards core network."""
         self._exec_command_in_bessd_workload(
-            command=f"ip route replace default via {self._get_core_network_gateway_ip_config()} metric 110"  # noqa: E501
+            command=f"ip route replace default via {self._get_network_gateway_ip_config('core')} metric 110"  # noqa: E501
```
patriciareinoso, line 620:
```suggestion
            command=f"ip route replace default via {self._get_network_gateway_ip_config(CORE_INTERFACE_NAME)} metric 110"  # noqa: E501
```

src/charm.py
```diff
@@ -620,84 +614,17 @@ def _create_bessctl_executed_validation_file(self, content) -> None:
...
         )
         logger.info("Default core network route created")
 
     def _create_ran_route(self) -> None:
         """Creates ip route towards gnb-subnet."""
         self._exec_command_in_bessd_workload(
-            command=f"ip route replace {self._get_gnb_subnet_config()} via {self._get_access_network_gateway_ip_config()}"  # noqa: E501
+            command=f"ip route replace {self._get_gnb_subnet_config()} via {self._get_network_gateway_ip_config('access')}"  # noqa: E501
```
patriciareinoso, line 627:
```suggestion
            command=f"ip route replace {self._get_gnb_subnet_config()} via {self._get_network_gateway_ip_config(ACCESS_INTERFACE_NAME)}"  # noqa: E501
```

patriciareinoso (changes requested):
Line 62
`SUPPORTED_UPF_MODES = ["af_packet", "dpdk"]`
can be replace by consts as well

patriciareinoso (approved):

gatici (approved):

config.yaml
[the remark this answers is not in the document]
dariofaccin: Added right now.

src/charm.py
dariofaccin: As discussed in these days, having the `self._charm_config` in the constructor may not be enough, as in some cases (`ops.Harness` during testing and deferred events in production) the charm is not reinstantiated. For this reason, I override the object in the `config-changed` and `pebble-ready` events for now.

src/charm_config.py
```diff
@@ -0,0 +1,148 @@
...
+    @field_validator("access_interface_mac_address", "core_interface_mac_address", mode="before")
+    @classmethod
+    def validate_interface_mac_address(cls, value: str, info: ValidationInfo) -> str:
+        """Validate that IP network address is valid."""
+        MacAddress.validate_mac_address(value.encode())
+        return value
+
+
+@dataclasses.dataclass(frozen=True)
+class CharmConfig:
```
Gmerold, line 119: How about adding a constructor to this class, so that we can return config options separately? If we had that, in the `charm.py`, instead of having a bunch of getters doing `_charm_config.upf_config.some_config_option` we could simply do `_charm_config.some_config_option`. What do you think?

Gmerold (commented):
Other than the question I left, looks good to me.

Gmerold (approved):

PR #86, opened 2024-02-02
fix: TELCO-996 Use blocked status instead of throwing exception

markbeierl:
# Description

Instead of using an exception, which results in messages that are not helpful to the user (ie: hook failed), this change puts the charm into blocked state instead.

# Checklist:

- N/A I have made corresponding changes to the documentation
- [X] I have added tests that validate the behaviour of the software
- [X] I validated that new and existing unit tests pass locally with my changes
- N/A Any dependent changes have been merged and published in downstream modules
- N/A I have bumped the version of the library

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -207,7 +207,7 @@ class UPFOperatorCharm(CharmBase):
         Args:
             event: Juju event
         """
-        if not self._cpu_check():
+        if not self._is_cpu_compatible():
             return
         self._create_external_upf_service()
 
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -509,15 +492,38 @@ def _on_config_changed(self, event: EventBase):
         self._update_fiveg_n3_relation_data()
         self._update_fiveg_n4_relation_data()
 
+    def _cpu_check(self):
```
gatici, line 495: I would suggest to give a name which is more descriptive such as `_is_cpu_check_passed`.

gatici (changes requested):

gatici (approved):

PR #108, opened 2024-02-29
feat: implement log forwarding

gruyaume:
# Description

Implement log forwarding via Pebble's new `log-target` spec. We use the `loki_push_api` charm relation to get the loki endpoint.

# Checklist:


markbeierl:
Do we want to create subtasks in JIRA for this so we can scale it out and track which ones are done?

markbeierl (approved):

gruyaume:
> Do we want to create subtasks in JIRA for this so we can scale it out and track which ones are done?

I don't know that this is necessary, it's quite a simple change and I'm ok making it in all our charms.

PR #116, opened 2024-03-05
fix: Using charm name as DEFAULT_FIELD_MANAGER

gatici:
# Description

The global constant called DEFAULT_FIELD_MANAGER is set to a default value which is `controller` refers the name of Juju controller that is not always correct.  Replacing  DEFAULT_FIELD_MANAGER with the charm name as it is the charm in charge of managing the field.

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -63,9 +63,6 @@ REQUIRED_CPU_EXTENSIONS = ["avx2", "rdrand"]
 REQUIRED_CPU_EXTENSIONS_HUGEPAGES = ["pdpe1gb"]
 LOGGING_RELATION_NAME = "logging"
 
-# The default field manager set when using kubectl to create resources
-DEFAULT_FIELD_MANAGER = "controller"
-
 
 class NadConfigChangedEvent(EventBase):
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -170,7 +167,7 @@ def _create_external_upf_service(self) -> None:
             ),
         )
 
-        client.apply(service, field_manager=DEFAULT_FIELD_MANAGER)
+        client.apply(service, field_manager=self.__class__.__name__)
```
Gmerold, line 170: While there's nothing wrong with this, did you consider using `self.app.name` here? 
I'm not sure if this will be visible to the user anywhere, but if yes, perhaps pointing to the application name would be slightly better than to the class name?
WDYT?

Gmerold (approved):
One small question.
Other than that LGTM.

src/charm.py
gatici: The user could see the manager by using --show-managed-fields option if desired. 
Using the app name makes sense to me.

Gmerold (approved):

PR #129, opened 2024-03-13
chore: handle charm status with ops CollectStatus event

dariofaccin:
# Description

This PR aims to use new ops event `CollectStatus` to manage status in the charm. After the event named "collect_unit_status" (triggered after every hook), the callback sets the status of charm automatically.

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -551,6 +551,8 @@ class UPFOperatorCharm(CharmBase):
         """Handle pfcp agent Pebble ready event."""
         if self._check_cpu_compatibility():
             return
+        if self._unit_is_non_active_status():
+            return
         self._configure_pfcp_agent_workload()
 
     def _configure_bessd_workload(self) -> None:
--- a/src/charm.py
+++ b/src/charm.py
@@ -31,15 +31,7 @@ from lightkube.core.client import Client
 from lightkube.models.core_v1 import ServicePort, ServiceSpec
 from lightkube.models.meta_v1 import ObjectMeta
 from lightkube.resources.core_v1 import Node, Pod, Service
-from ops import (
-    ActiveStatus,
-    BlockedStatus,
-    Container,
-    ModelError,
-    RemoveEvent,
-    StatusBase,
-    WaitingStatus,
-)
+from ops import ActiveStatus, BlockedStatus, Container, ModelError, RemoveEvent, WaitingStatus
 from ops.charm import CharmBase, CharmEvents, CollectStatusEvent
 from ops.framework import EventBase, EventSource
 from ops.main import main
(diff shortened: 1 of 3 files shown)


tests/integration/test_integration.py
```diff
@@ -28,7 +28,6 @@ async def _deploy_grafana_agent(ops_test: OpsTest):
 
 
 @pytest.fixture(scope="module")
-@pytest.mark.abort_on_fail
```
Gmerold, line 31: Why do want to remove this?

src/charm.py
```diff
@@ -86,14 +93,6 @@ class UPFOperatorCharm(CharmBase):
...
     def __init__(self, *args):
         super().__init__(*args)
-        if not self.unit.is_leader():
-            # NOTE: In cases where leader status is lost before the charm is
-            # finished processing all teardown events, this prevents teardown
-            # event code from running. Luckily, for this charm, none of the
-            # teardown code is necessary to perform if we're removing the
-            # charm.
-            self.unit.status = BlockedStatus("Scaling is not implemented for this charm")
-            return
```
gruyaume, line 96: shouldn't we still return? Here removing this means that we will observe events for non leaders. I.e. it's a change of behavior from the current state of things

src/charm.py
```diff
@@ -228,7 +228,7 @@ def _on_install(self, event: EventBase) -> None:
         Args:
             event: Juju event
         """
-        if not self._is_cpu_compatible():
+        if self._check_cpu_compatibility():
```
Gmerold, line 231: I liked the previous name better. IMO it reads better.

src/charm.py
```diff
@@ -892,22 +905,19 @@ def _is_cpu_compatible(self) -> bool:
             required_extension in self._get_cpu_extensions()
             for required_extension in REQUIRED_CPU_EXTENSIONS
         ):
-            self.unit.status = BlockedStatus(
+            return BlockedStatus(
```
Gmerold, line 908: This seems like a one hell of a long status message to be printed in a Juju status screen. Does it even fit the table nicely?
How about we keep the return type of the whole method as a `bool`, log the error message rather than print it in the status, set the `Blocked` status in the `_on_collect_unit_status` if this returns `False` and add a generic status message?
WDYT?

Gmerold: BTW, if we go with the current impl, docstring needs to be updated.

src/charm.py
```diff
@@ -487,66 +487,94 @@ def _hwcksum_config_matches_pod_config(self) -> bool:
...
+        if not self._bessd_container.can_connect():
+            return WaitingStatus("Waiting for bessd container to be ready")
+        if not self._kubernetes_multus.is_ready():
+            return WaitingStatus("Waiting for Multus to be ready")
+        if not self._bessd_container.exists(path=BESSD_CONTAINER_CONFIG_PATH):
+            return WaitingStatus("Waiting for storage to be attached")
+        if status := self._check_container_availability():
+            return status
+
+    def _check_container_availability(self):
```
Gmerold, line 515: I think the name of the method doesn't really match the implementation.

src/charm.py
Gmerold: I'm just wondering - isn't it a good case to actually use `defer`? I mean if the CPU is not compatible, there's no way to recover from that. If we defer this even, it will always run before any other events and block the charm. Unless I'm missing something, this seems like a simpler solution comparing to what we have now.
WDYT?

Gmerold (changes requested):

src/charm.py
```diff
@@ -487,66 +487,94 @@ def _hwcksum_config_matches_pod_config(self) -> bool:
...
         """Handler for config changed events."""
-        # workaround for https://github.com/canonical/operator/issues/736
-        try:
+        try:  # workaround for https://github.com/canonical/operator/issues/736
             self._charm_config: CharmConfig = CharmConfig.from_charm(charm=self)  # type: ignore[no-redef]  # noqa: E501
-        except CharmConfigInvalidError as exc:
-            self.model.unit.status = BlockedStatus(exc.msg)
+        except CharmConfigInvalidError:
             return
         if not self.unit.is_leader():
```
gatici, line 536:
``` if not self.unit.is_leader():
            return
``` 
Could we do this check in the charm constructor instead of here ?

src/charm.py
```diff
@@ -487,66 +487,94 @@ def _hwcksum_config_matches_pod_config(self) -> bool:
...
+        try:  # workaround for https://github.com/canonical/operator/issues/736
+            self._charm_config: CharmConfig = CharmConfig.from_charm(charm=self)  # type: ignore[no-redef]  # noqa: E501
+        except CharmConfigInvalidError as exc:
+            return BlockedStatus(exc.msg)
+        if status := self._check_cpu_compatibility():
+            return status
+        if not self._kubernetes_multus.multus_is_available():
+            return BlockedStatus("Multus is not installed or enabled")
+        if not self._bessd_container.can_connect():
+            return WaitingStatus("Waiting for bessd container to be ready")
```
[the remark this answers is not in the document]
dariofaccin: I created this PR only to migrate to the CollectStatus event. I'd prefer handling the `ConnectionError` issue in another PR, as it may involve adding different sanity checks.

tests/integration/test_integration.py
dariofaccin: It does nothing, according to the [documentation](https://docs.pytest.org/en/stable/deprecations.html#applying-a-mark-to-a-fixture-function). We already have the marker on the actual test.

src/charm.py
dariofaccin: What about returning only two different messages (one for CPU not compatible, one for not HugePages available)? Otherwise we lose information on whether the blocked status depends on CPU or HugePages.

src/charm.py
dariofaccin: We don't have any guarantee that the unit that was leader when the constructor is instantiated is still leader when config-changed event is triggered. From [SDK](https://juju.is/docs/sdk/leadership):

> Units that hold leader status should not assume they will retain it, a new leader can be elected at any time.

src/charm.py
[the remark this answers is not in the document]
Gmerold: Yeah, we've agreed on the same with Gulsum. I'll bring it up on the daily today.

src/charm.py
Gmerold: We don't lose the information - what I was thinking was logging the exact reason:
```python
logger.error("The exact error desc goes here")
return False
```
And then in the `_unit_is_non_active_status`, something along the lines of:
```python
if not self._is_cpu_compatible():
    return BlockedStatus("Incompatible CPU. Check logs for details.")
```

src/charm.py
```diff
@@ -487,66 +483,109 @@ def _hwcksum_config_matches_pod_config(self) -> bool:
...
+        if not self._kubernetes_multus.is_ready():
+            logger.info("Waiting for Multus to be ready")
+            return WaitingStatus("Waiting for Multus to be ready")
+        if not self._bessd_container.exists(path=BESSD_CONTAINER_CONFIG_PATH):
+            logger.info("Waiting for storage to be attached")
+            return WaitingStatus("Waiting for storage to be attached")
+        if container_status := self._container_status_setter():
+            return container_status
+
+    def _container_status_setter(self):
```
gruyaume, line 517: This method lacks return typing plus it's name could be better. It is not a setter.

src/charm.py
```diff
@@ -487,66 +491,122 @@ def _hwcksum_config_matches_pod_config(self) -> bool:
...
+            logger.info("Waiting for Multus to be ready")
+            return WaitingStatus("Waiting for Multus to be ready")
+        if not self._bessd_container.exists(path=BESSD_CONTAINER_CONFIG_PATH):
+            logger.info("Waiting for storage to be attached")
+            return WaitingStatus("Waiting for storage to be attached")
+        if container_status := self._container_services_are_running():
+            return container_status
+        return None
+
+    def _container_services_are_running(self) -> Optional[WaitingStatus]:
```
gatici, line 531: I would suggest to give a name ` _unavailable_container_services` as it returns not running services.

gruyaume (approved):

Gmerold (approved):

PR #140, opened 2024-03-27
feat: Publish charm to branch on push

markbeierl:


.github/workflows/main.yaml
```diff
@@ -54,10 +56,11 @@ jobs:
     if: ${{ (github.ref_name != 'main') && (github.event_name == 'push') }}
     uses: canonical/sdcore-github-workflows/.github/workflows/publish-charm.yaml@main
     with:
-      branch-name: /${{ github.ref_name }}
+      branch-name: ${{ github.ref_name }}
```
gatici, line 59: Need to remove main branch from pull_requests (line 6)

markbeierl: No, we need to keep running the same jobs on pull request. This change is only for pushes and does not impact the PR workflow.

gatici (approved):

PR #146, opened 2024-04-02
chore: Update configuration logic

markbeierl:
# Description

Changes the bessd config check from file presence based to query process state.

# Checklist:

- [X] My code follows the [style guidelines](/CONTRIBUTING.md) of this project
- [X] I have performed a self-review of my own code
N/A I have made corresponding changes to the documentation
- [X] I validated that new and existing unit tests pass locally with my changes
N/A Any dependent changes have been merged and published in downstream modules
N/A I have bumped the version of the library

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -669,7 +669,7 @@ class UPFOperatorCharm(CharmBase):
             logger.info("Service `bessd` configuration script complete")
             for line in stdout.splitlines():
                 logger.debug("`up4.bess`: %s", line)
-            if isinstance(stderr, str) and len(stderr) > 0:
+            if stderr:
                 for line in stderr.split():
                     logger.error("`up4.bess`: %s", line)
             return
--- a/src/charm.py
+++ b/src/charm.py
@@ -725,7 +725,7 @@ class UPFOperatorCharm(CharmBase):
             logger.debug("bessd configured workers: %s", stdout)
             return True
         except ExecError as e:
-            logger.info(f"Configuration check: {e}")
+            logger.info("Configuration check: %s", e)
             return False
 
     def _configure_bessd_for_dpdk(self) -> None:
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -640,65 +649,95 @@ def _configure_bessd_workload(self) -> None:
...
+            (stdout, stderr) = self._exec_command_in_bessd_workload(
+                command=command,
+                environment=self._bessd_environment_variables,
+                timeout=30,
+            )
+
+            logger.info("Service `bessd` configuration script complete")
+            for line in stdout.splitlines():
+                logger.debug("`up4.bess`: %s", line)
+            if isinstance(stderr, str) and len(stderr) > 0:
```
ghislainbourgeois, line 672: The return value of `_exec_command_in_bessd_workload` is a tuple of strings, so `stderr` should always be a string. This line could be simplified to this:
```suggestion
            if stderr:
```

src/charm.py
```diff
@@ -640,65 +649,95 @@ def _configure_bessd_workload(self) -> None:
...
+        command = "/opt/bess/bessctl/bessctl show worker"
+        try:
+            (stdout, _) = self._exec_command_in_bessd_workload(
+                command=command,
+                timeout=10,
+            )
+            logger.debug("bessd configured workers: %s", stdout)
+            return True
+        except ExecError as e:
+            logger.info(f"Configuration check: {e}")
```
ghislainbourgeois, line 728: The code style guide disallows f-strings in logging calls.
```suggestion
            logger.info("Configuration check: %s", e)
```

ghislainbourgeois (changes requested):
Looks good, only a couple of small comments.

src/charm.py
markbeierl: One of the linters was complaining about it in the k8s charm, and I had to use that syntax there. I'll try it to see if the signature is different here and will pass

ghislainbourgeois: Ok, if this suggestion does not work let me know, this is not a big concern.

ghislainbourgeois (approved):

PR #147, opened 2024-04-03
refactor: use later `begin` rather than re-instantiating the harness charm 

tonyandrewmeyer:
# Description

The charm conditionally sets the unit status (and does some other work, including observing) based on the result of setting up a configuration object during `__init__`. This isn't a practice that Charm Tech recommends, and it's one that currently works in production (if defer is not used) but is not supported by Harness (which [won't be changed, unfortunately](https://github.com/canonical/operator/issues/736)).

At the moment, the tests have a workaround where the harness charm is reinstantiated using private methods of both the harness framework and harness itself. This PR removes the reinstantiation, so that the private use is not required.

The actual change is pretty straightforward: it just waits until all the setup is complete (in particular, setting the config values) before calling `harness.begin()` (which executes the charm's `__init__`).

This PR does not change any of the tests, other than replacing the `reinstantiate_charm()` call with `self.harness.begin()`. To minimise the impact of the change, these tests are moved to a separate `TestCase` subclass. An alternative that would result in fewer lines changing (e.g. in blame) would be to remove the `begin()` from the existing class and add it in explicitly to each test case (that doesn't do the `reinstantiate()`), but this change seemed cleaner.

# Checklist:

- [ ] I have made corresponding changes to the documentation - N/A
- [ ] I have added tests that validate the behaviour of the software - N/A
- [ ] Any dependent changes have been merged and published in downstream modules - N/A
- [ ] I have bumped the version of the library - N/A

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -72,22 +72,7 @@ def update_nad_labels(nads: list[NetworkAttachmentDefinition], app_name: str) ->
         nad.metadata.labels = {"app.juju.is/created-by": app_name}
 
 
-class TestCharm(unittest.TestCase):
-
-    def reinstantiate_charm(self):
-        charm = self.harness.charm
-        self.harness.framework._forget(charm)
-        self.harness.framework._forget(charm.on)
-        self.harness.framework._forget(charm.fiveg_n3_provider)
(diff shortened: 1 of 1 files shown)


ghislainbourgeois (approved):

ghislainbourgeois:
Thanks for the PR and the detailed explanation!

PR #154, opened 2024-04-11
chore: Move to codespell and ruff 

patriciareinoso:
# Description

Move to ruff and codespell for linting. `pyproject-flake8` is currently broken on Python 3.12, making development on Noble break.

# Checklist:

### DIFF
--- a/lib/charms/sdcore_upf_k8s/v0/fiveg_n3.py
+++ b/lib/charms/sdcore_upf_k8s/v0/fiveg_n3.py
@@ -93,7 +93,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
 
 PYDEPS = ["pydantic", "pytest-interface-tester"]
 
@@ -129,7 +129,7 @@ class ProviderSchema(DataBagSchema):
 
 
 def data_matches_provider_schema(data: dict) -> bool:
-    """Returns whether data matches provider schema.
+    """Return whether data matches provider schema.
 
     Args:
         data (dict): Data to be validated.
@@ -149,18 +149,18 @@ class FiveGN3RequestEvent(EventBase):
     """Dataclass for the `fiveg_n3` request event."""
 
     def __init__(self, handle, relation_id: int):
-        """Sets relation id."""
+        """Set relation id."""
         super().__init__(handle)
         self.relation_id = relation_id
 
     def snapshot(self) -> dict:
-        """Returns event data."""
+        """Return event data."""
         return {
             "relation_id": self.relation_id,
         }
 
     def restore(self, snapshot):
-        """Restores event data."""
+        """Restore event data."""
         self.relation_id = snapshot["relation_id"]
 
 
@@ -176,7 +176,7 @@ class N3Provides(Object):
     on = N3ProviderCharmEvents()
 
     def __init__(self, charm: CharmBase, relation_name: str):
-        """Observes relation joined event.
+        """Observe relation joined event.
 
         Args:
             charm: Juju charm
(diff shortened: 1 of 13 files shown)


pyproject.toml
```diff
@@ -9,24 +9,29 @@ show_missing = true
...
+    "D400",
+    "D404",
+    "D406",
+    "D407",
+    "D408",
+    "D409",
+    "D413",
+]
+ignore = ["E501", "D107"]
+per-file-ignores = {"tests/*" = ["D100","D101","D102","D103","D104"]}
```
Gmerold, line 31: Seems like we're ignoring much than we used to. Is there any reason for that?

pyproject.toml
```diff
@@ -9,24 +9,29 @@ show_missing = true
...
-exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv"]
-select = ["E", "W", "F", "C", "N", "R", "D", "H"]
-# Ignore D107 Missing docstring in __init__
-ignore = ["D107"]
-# D100, D101, D102, D103: Ignore missing docstrings in tests
-per-file-ignores = ["tests/*:D100,D101,D102,D103,D104"]
-docstring-convention = "google"
+
+[tool.codespell]
+skip = "build,lib,venv,icon.svg,.tox,.git,.mypy_cache,.ruff_cache,.coverage"
```
Gmerold, line 37: Do we really want to skip `lib` completely? I think that's fine for the external libs, but this will also affect the libs provided by this charm, right?

lib/charms/sdcore_upf_k8s/v0/fiveg_n3.py
```diff
@@ -216,16 +216,16 @@ class N3AvailableEvent(EventBase):
...
         self.upf_ip_address = upf_ip_address
 
     def snapshot(self) -> dict:
-        """Returns event data."""
+        """Return event data."""
         return {"upf_ip_address": self.upf_ip_address}
 
     def restore(self, snapshot):
-        """Restores event data."""
+        """Restore event data."""
```
Gmerold, line 228: I assume that's because of the new tool, correct?

Gmerold (changes requested):

pyproject.toml
patriciareinoso: we were using the google convention.
The Google convention includes all D errors apart from those on the list. 
I just found out the there is another way of enforcing the convention
```
[tool.ruff.lint.pydocstyle]
convention = "google"
```

lib/charms/sdcore_upf_k8s/v0/fiveg_n3.py
patriciareinoso: yes, indeed

pyproject.toml
patriciareinoso: you are rigth, this is sking the libs that belong to the charm. We are already filtering external libs in tox.ini when setting `lib_path` to the libs that belong to the charm
i will fix it.

pyproject.toml
Gmerold: OK.
What about `E501`? We're setting the desired line length and then we ignore the violations?

patriciareinoso: True, I think that we can disable it only from tests. Like this we can remove all the #noqa
What do you think ?

pyproject.toml
patriciareinoso: actually we are only filtering for ruff but not for codespell. I need to check this because if I remove it, it will fail if an external lib has a misspelling error

patriciareinoso: fixed. Codespell is executed on the charm's libs on the tox.ini file

pyproject.toml
patriciareinoso: I re enabled the line length again

Gmerold (approved):

PR #220, opened 2024-05-15
fix: route_control not being able to add routes to accessRoutes and coreRoutes modules

Gmerold:
# Description

CONTEXT:
In Charmed Aether SD-Core E2E we would often observe failures caused by the `route_control` service not being able to add routes to `accessRoutes` and `coreRoutes` modules. It happens when the modules are not present in `bess` at the time of starting the `route_control`. The modules are expected to be created by the `bess` configuration script - `up4.bess`. To create modules, `up4.bess` script requires certain ARP entries to be present. It may happen, that the required entries are not yet present in the ARP table when the `bess` configuration script runs. In that case, it will exit cleanly without creating the required modules.

This PR addresses the problem by introducing following changes to the UPF charm startup procedure:
- adds `accessRoutes` and `coreRoutes` modules existence checks before assuming that `bess` is configured
- separates Pebble layers for `bessd` and `routectl` services, which makes starting `routectl` dependent of successful configuration of `bess`

TESTS:
This change has been tested by building a custom charm and running E2E tests against it. The results are available [here](https://github.com/canonical/sdcore-tests/actions/runs/9096979677). Run `#10` failed because of the `hook failed: "logging-relation-changed" for grafana-agent:logging-provider` which is not related to this PR.

ADDITIONAL CHANGES:
Additionally, the PR changes the `bess configuration file` phrase to `UPF configuration file` when talking about the `upf.json`

# Checklist:


src/charm.py
```diff
@@ -714,19 +750,36 @@ def _is_bessd_grpc_service_ready(self) -> bool:
     def _is_bessd_configured(self) -> bool:
```
gruyaume, line 750: Should we rename this method to `_is_bessd_running`?

src/charm.py
```diff
@@ -714,19 +750,36 @@ def _is_bessd_grpc_service_ready(self) -> bool:
...
         """Check if bessd has been configured.
 
-        Examine the output from bessctl to show worker. If there is no
-        active worker, bessd is assumed not to be configured.
+        Bess is considered as configured when the `accessRoutes` and `coreRoutes` modules
+        are available and at least one worker is in the RUNNING state. This method examines
+        the output of the `bessctl show` command to determine whether the above criteria are met.
```
gruyaume, line 755: Good job on the docstrings overall in this PR, they add useful context

src/charm.py
```diff
@@ -714,19 +750,36 @@ def _is_bessd_grpc_service_ready(self) -> bool:
...
 
         Returns:
-            bool:   True/False
+            bool: Whether the `bessd` service is configured
         """
-        command = "/opt/bess/bessctl/bessctl show worker"
+        show_accessRoutes_module_cmd = "/opt/bess/bessctl/bessctl show module accessRoutes"  # noqa: N806
+        show_coreRoutes_module_cmd = "/opt/bess/bessctl/bessctl show module coreRoutes"  # noqa: N806
+        show_worker_cmd = "/opt/bess/bessctl/bessctl show worker"
         try:
```
gruyaume, line 763: There is quite a lot inside the same try/except, should we separate those in smaller try/excepts to better catch exceptions?

gruyaume (commented):
Overall this looks good!

src/charm.py
Gmerold: As discussed, since all these commands throw the same exception I think it makes sense to keep them under one `try-except` block. We also log the exact error message, so we'll know which check failed.
One thing I've changed is the log level for the error message. It's now `error`, so it'll be clearly visible in the debug-log.

gatici (approved):

gruyaume (approved):

src/charm.py
```diff
@@ -714,22 +750,39 @@ def _is_bessd_grpc_service_ready(self) -> bool:
...
+            logger.debug(
+                "bessd configured accessRoutes module: %s", show_accessRoutes_module_stdout
+            )
+            (show_coreRoutes_module_stdout, _) = self._exec_command_in_bessd_workload(  # noqa: N806
+                command=show_coreRoutes_module_cmd,
+                timeout=10,
+            )
+            logger.debug("bessd configured coreRoutes module: %s", show_coreRoutes_module_stdout)
+            (show_worker_stdout, _) = self._exec_command_in_bessd_workload(
+                command=show_worker_cmd,
```
markbeierl, line 777: I think this could be the first check. If there are no workers, there should not be any modules.

src/charm.py
```diff
@@ -714,22 +750,39 @@ def _is_bessd_grpc_service_ready(self) -> bool:
...
                 timeout=10,
             )
-            logger.debug("bessd configured workers: %s", stdout)
+            if "RUNNING" not in show_worker_stdout:
+                return False
+            logger.debug("bessd configured workers: %s", show_worker_stdout)
             return True
         except ExecError as e:
-            logger.info("Configuration check: %s", e)
+            logger.error("Configuration check failed: %s", e)
```
markbeierl, line 785: Also the reason why this was not written as error before, is that it is normal for the first check to fail. If the up4 script has not run, this is our indication that it needs to run.

src/charm.py
Gmerold: Done.

src/charm.py
Gmerold: I'd still have it as `error`. The problem I have with the `info` logs in such cases is that it's easy to miss them. Errors are printed in red, so it's super easy to find them.

PR #474, opened 2025-03-11
feat: add ip masquerading configuration option

gruyaume:
# Description

Here we add the `core-ip-masquerade` configuration option, allowing users to decide whether to use NATting for user equipment traffic. When set to `true`, the UPF will perform NATting for User Equipment traffic, packets will have the `core-ip` address as their source IP address. If not set, User Equipment packets will keep their original IP addresses.

## This is not a breaking change

The default behavior is the same: The UPF performs NATting using the `core-ip` address. But we now allow to switch it off. 
 
# Checklist:


patriciareinoso (changes requested):
`core_ip_masquerade` is sometimes a string and sometimes a bool. It can get confusing. I think we should make the difference in naming

src/charm.py
```diff
@@ -666,12 +666,15 @@ def _create_upf_configuration_file(self) -> tuple[bool, bool]:
...
         recreate_pod = False
         core_ip_address = self._get_network_ip_config(CORE_INTERFACE_NAME)
+        core_ip_masquerade_flag = self._charm_config.core_ip_masquerade
         content = render_bessd_config_file(
             upf_hostname=self._upf_hostname,
             upf_mode=self._charm_config.upf_mode,
             access_interface_name=ACCESS_INTERFACE_NAME,
             core_interface_name=CORE_INTERFACE_NAME,
-            core_ip_address=core_ip_address.split("/")[0] if core_ip_address else "",
+            core_ip_masquerade=core_ip_address.split("/")[0]
```
ghislainbourgeois, line 675:
```suggestion
            core_ip_address=core_ip_address.split("/")[0]
```

I would rename this to `core_ip_address` for clarity. It is what it contains after all.

src/charm.py
```diff
@@ -666,12 +666,15 @@ def _create_upf_configuration_file(self) -> tuple[bool, bool]:
         restart = False
         recreate_pod = False
         core_ip_address = self._get_network_ip_config(CORE_INTERFACE_NAME)
+        core_ip_masquerade_flag = self._charm_config.core_ip_masquerade
```
ghislainbourgeois, line 669:
```suggestion
        core_ip_masquerade_enabled = self._charm_config.core_ip_masquerade
```

I think this would a bit clearer with this name. Maybe the charm configuration itself could be renamed like that, WDYT?

ghislainbourgeois (changes requested):

src/charm.py
gruyaume: Ok I addressed this. I initially went for `core_ip_masquerade` because this is the variable used in the json configuration file.

src/charm.py
gruyaume: Agreed, fixed

gruyaume:
> `core_ip_masquerade` is sometimes a string and sometimes a bool. It can get confusing. I think we should make the difference in naming

Ok I addressed this, let me know what you think.

ghislainbourgeois (approved):

PR #484, opened 2025-04-17
feat: add configuration to start BESS HTTP server

dariofaccin:
# Description

This PR adds a new configuration option, `bess-http` to the UPF charm. The option accepts boolean values.
When enabled, the UPF will start and expose the bess HTTP server on port 5000.
A new service `besstl-http` is added to the `bessd` container plan. By default, the service [startup](https://documentation.ubuntu.com/pebble/reference/layer-specification/) is `disabled`.
When the configuration option is changed, the following operations are performed:
* if the new service startup value is `enabled` and the current status is `inactive`, the service is started
* if the new service startup value is `disabled` and the current status is `active`, the service is stopped

This logic is overridden when a service restart is forced by the caller of the service configuration function (the behaviour is unchanged).

# Checklist:

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -180,3 +180,4 @@ config:
       default: false
       description: |
         When enabled, UPF will expose the BESS HTTP server on port 5000.
+        WARNING: This should be enabled only for debugging purposes and disabled afterwards.
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -175,3 +175,8 @@ config:
       type: string
       default: info
       description: Log level for the UPF. One of `debug`, `info`, `warn`, `error`, `fatal`, `panic`.
+    bess-http:
+      type: boolean
+      default: false
+      description: |
+        When enabled, UPF will expose the BESS HTTP server on port 5000
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -175,8 +175,8 @@ config:
       type: string
       default: info
       description: Log level for the UPF. One of `debug`, `info`, `warn`, `error`, `fatal`, `panic`.
-    bess-http:
+    enable-bess-http:
       type: boolean
       default: false
       description: |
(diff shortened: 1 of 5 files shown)


charmcraft.yaml
```diff
@@ -175,3 +175,8 @@ config:
       type: string
       default: info
       description: Log level for the UPF. One of `debug`, `info`, `warn`, `error`, `fatal`, `panic`.
+    bess-http:
```
patriciareinoso, line 178:
```suggestion
    enable-bess-http:
```
can we rename the config parameter to something like this ? it would be more evident that it's a boolean parameter.

charmcraft.yaml
```diff
@@ -175,3 +175,8 @@ config:
...
+      type: boolean
+      default: false
+      description: |
+        When enabled, UPF will expose the BESS HTTP server on port 5000
```
ghislainbourgeois: I would add something in the vein of:

```
WARNING: This should be enabled only for debugging and disabled afterwards.
```

I want to make it clear that it's not just a nice feature that one enables on their production deployment.

src/charm.py
```diff
@@ -705,12 +706,20 @@ def _create_and_configure_bessd_service(self, restart_service: bool):
         (e.g. when the UPF configuration file has changed).
         """
         plan = self._bessd_container.get_plan()
-        if not all(service in plan.services for service in self._bessd_pebble_layer.services):
+        if not all(service in plan.services == service for
```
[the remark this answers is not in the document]
ghislainbourgeois: I agree it is not very clear, we should find a clearer way to express this condition.

ghislainbourgeois (changes requested):
Looks, good, I left 2 small comments.

On another note, once this is merged, we should update the troubleshooting documentation to mention this feature and how to use it.

ghislainbourgeois (approved):

gatici (approved):

